PR #15, opened 2020-11-13
feat: Add support for Unleash Context properties

### LINKED ISSUE #3: Bug: Custom "Properties" object not supported

### LINKED ISSUE #14: IContext doesn't implement unleash context
IContext interface right now is an object with string keys and string values. It should implement the Unleash Context defined here: https://www.unleash-hosted.com/docs/the-unleash-context/

mattmurph9:
Fixes #14 
Now, the IContext is an object of arbitrary key value pairs. Extend this to implement the Unlease Context (https://www.unleash-hosted.com/docs/the-unleash-context)

and fixes #3 
We want to allow a consumer of the UnleashClient to be able to send the properties field, as mentioned in the Unlease Context. This allows them to support custom activation strategies. In this PR I flatten the properties into the url search params, so a properties object like
```
properties: {
  property1: 'property1Value',
  property2: 'property2Value'
}
```
would yield url search params `properties.property1=property1Value&properties.property2=property2Value`

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,6 +1,6 @@
 import { FetchMock } from 'jest-fetch-mock';
 import * as data from '../tests/example-data.json'; 
-import { EVENTS, IConfig, UnleashClient } from './index';
+import { EVENTS, IConfig, IMutableContext, UnleashClient } from './index';
 
 jest.useFakeTimers();
 
@@ -172,6 +172,44 @@ test('Should include context fields on request', async () => {
         [JSON.stringify(data), { status: 200 }],
         [JSON.stringify(data), { status: 304 }],
     );
+    const context: IMutableContext = {
+        userId: '123',
+        sessionId: '456',
+        remoteAddress: 'address',
+        properties: {
+            property1: 'property1',
+            property2: 'property2'
+        }
+    }
+    const config: IConfig = {
+        url: 'http://localhost/test',
+        clientKey: '12',
+        appName: 'web',
+        environment: 'prod',
+        context
+    };
+    const client = new UnleashClient(config);
+
+    await client.start();
+
+    jest.advanceTimersByTime(1001);
+
+    const url = new URL(fetchMock.mock.calls[0][0]);
+
+    expect(url.searchParams.get('userId')).toEqual('123');
+    expect(url.searchParams.get('sessionId')).toEqual('456');
+    expect(url.searchParams.get('remoteAddress')).toEqual('address');
+    expect(url.searchParams.get('properties.property1')).toEqual('property1');
+    expect(url.searchParams.get('properties.property2')).toEqual('property2');
+    expect(url.searchParams.get('appName')).toEqual('web');
+    expect(url.searchParams.get('environment')).toEqual('prod');
+});
+
+test('Should update context fields on request', async () => {
+    fetchMock.mockResponses(
+        [JSON.stringify(data), { status: 200 }],
+        [JSON.stringify(data), { status: 304 }],
+    );
     const config: IConfig = {
         url: 'http://localhost/test',
         clientKey: '12',
@@ -213,18 +251,20 @@ test('Should not add property fields when properties is an empty object', async 
         url: 'http://localhost/test',
         clientKey: '12',
         appName: 'web',
-        environment: 'prod'
+        environment: 'prod',
+        context: {
+            properties: {}
+        }
     };
     const client = new UnleashClient(config);
-    client.updateContext({
-        properties: {}
-    });
 
     await client.start();
 
     jest.advanceTimersByTime(1001);
 
     const url = new URL(fetchMock.mock.calls[0][0]);
+
+    console.log(url.toString(), url.searchParams.toString(), url.searchParams.get('properties'));
 
     expect(url.searchParams.get('appName')).toEqual('web');
     expect(url.searchParams.get('environment')).toEqual('prod');
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -204,6 +204,33 @@ test('Should include context fields on request', async () => {
     expect(url.searchParams.get('environment')).toEqual('prod');
 });
 
+test('Should not add property fields when properties is an empty object', async () => {
+    fetchMock.mockResponses(
+        [JSON.stringify(data), { status: 200 }],
+        [JSON.stringify(data), { status: 304 }],
+    );
+    const config: IConfig = {
+        url: 'http://localhost/test',
+        clientKey: '12',
+        appName: 'web',
+        environment: 'prod'
+    };
+    const client = new UnleashClient(config);
+    client.updateContext({
+        properties: {}
+    });
+
+    await client.start();
+
+    jest.advanceTimersByTime(1001);
+
+    const url = new URL(fetchMock.mock.calls[0][0]);
+
+    expect(url.searchParams.get('appName')).toEqual('web');
+    expect(url.searchParams.get('environment')).toEqual('prod');
+    expect(url.searchParams.get('properties')).toBeNull();
+});
+
 test('Should use default environment', async () => {
     fetchMock.mockResponses(
         [JSON.stringify(data), { status: 200 }],
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -172,9 +172,22 @@ test('Should include context fields on request', async () => {
         [JSON.stringify(data), { status: 200 }],
         [JSON.stringify(data), { status: 304 }],
     );
-    const config: IConfig = { url: 'http://localhost/test', clientKey: '12', appName: 'web', environment: 'prod' };
+    const config: IConfig = {
+        url: 'http://localhost/test',
+        clientKey: '12',
+        appName: 'web',
+        environment: 'prod'
+    };
     const client = new UnleashClient(config);
-    client.updateContext({userId: '123', randomField: 'random'});
+    client.updateContext({
+        userId: '123',
+        sessionId: '456',
+        remoteAddress: 'address',
+        properties: {
+            property1: 'property1',
+            property2: 'property2'
+        }
+    });
 
     await client.start();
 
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -59,7 +72,11 @@ export class UnleashClient extends TinyEmitter {
             metricsInterval = 30,
             disableMetrics = false,
             environment = 'default',
-            appName}
+            appName,
+            userId,
+            sessionId,
+            remoteAddress,
+            properties}
```
mattmurph9, line 79: As of now these have no default value. This means when using them later, they could be undefined. I did this in case defaulting them to an empty string would overwrite the defaults in the pipeline somewhere.

src/index.ts
```diff
@@ -77,7 +94,7 @@ export class UnleashClient extends TinyEmitter {
         this.clientKey = clientKey;
         this.storage = storageProvider || new LocalStorageProvider();
         this.refreshInterval = refreshInterval * 1000;
-        this.context = {environment, appName};
+        this.context = {environment, appName, userId, sessionId, remoteAddress, properties};
```
mattmurph9, line 97: Since userId, sessionId, ... could be undefined, should they be conditionally added to the context?

src/index.ts
```diff
@@ -145,7 +162,18 @@ export class UnleashClient extends TinyEmitter {
             try {
                 const context = this.context;
                 const urlWithQuery = new URL(this.url.toString());
-                Object.keys(context).forEach((key) => urlWithQuery.searchParams.append(key, context[key]));
+                // Add context information to url search params. If the properties
+                // object is included in the context, flatten it into the search params
+                // e.g. /?...&property.param1=param1Value&property.param2=param2Value
+                Object.entries(context).forEach((contextEntry) => {
```
mattmurph9, line 168: I switched from Object.keys to Object.entries because I feel that it's cleaner. Less object accesses. Also saves some non-null assertions I believe. Can definitely switch back if necessary!

src/index.ts
```diff
@@ -8,14 +8,27 @@ export interface IConfig {
     clientKey: string;
     appName: string;
     environment?: string;
+    userId?: string;
```
ivarconr, line 11: it feels a bit cumbersome to have all context fields also defined as part of the config. The initial idea was that we have a few "static context fields" (`appName` and `envrionment`) which will never change, whereas the other context fields is more dynamic and can change in the lifetime of an application (e.g.: a user may login without a reload, or may be part of a new group after some actions). 

Maybe we could have a IStaticContext interface interface and have IContext inherit that and have a property on the config interface for the static context fields? What do you think?

mattmurph9: Ah absolutely I agree. It does feel cumbersome having all the context fields there. Let me clean it up a bit with these suggestions.

src/index.ts
frikkf: I think the context can have these static keys, but they should be omitted when parsed into the query params?

src/index.ts
frikkf, line 168: For better readability we could change contextEntry to be: ([contextKey, contextValue]) => ...

src/index.test.ts
```diff
@@ -183,7 +196,10 @@ test('Should include context fields on request', async () => {
     const url = new URL(fetchMock.mock.calls[0][0]);
 
     expect(url.searchParams.get('userId')).toEqual('123');
-    expect(url.searchParams.get('randomField')).toEqual('random');
+    expect(url.searchParams.get('sessionId')).toEqual('456');
+    expect(url.searchParams.get('remoteAddress')).toEqual('address');
+    expect(url.searchParams.get('properties.property1')).toEqual('property1');
+    expect(url.searchParams.get('properties.property2')).toEqual('property2');
```
frikkf, line 202: Would be great to have a use case where properties = {} (an empty object)

src/index.ts
mattmurph9: Great point!

src/index.test.ts
mattmurph9: Is there a case for properties being an empty object? Would the properties field be set at all if they were empty?

frikkf: A user of the client can do whatever they want, so I would expect an empty properties object to avoid adding query params.

src/index.ts
mattmurph9: @ivarconr see e16704a. This would be a breaking change though and require the consumer to define the Unleash Context and pass it in. Might not be what we want, but I wanted to ask your opinion. I think the types are a lot cleaner, but this we might want to keep the same UnleaseClient api.

src/index.test.ts
```diff
@@ -183,17 +195,46 @@ test('Should include context fields on request', async () => {
...
+
+    await client.start();
+
+    jest.advanceTimersByTime(1001);
+
+    const url = new URL(fetchMock.mock.calls[0][0]);
+
     expect(url.searchParams.get('appName')).toEqual('web');
     expect(url.searchParams.get('environment')).toEqual('prod');
+    expect(url.searchParams.get('properties')).toBeNull();
```
frikkf, line 229: Should it be null or undefined?

src/index.ts
```diff
@@ -3,20 +3,32 @@ import Metrics from './metrics';
...
+    remoteAddress?: string;
+    properties?: {
+        [key:string]: string,
+    };
+}
+
+export type IContext = IStaticContext & IMutableContext;
+
 export interface IConfig {
+    context: IContext;
```
frikkf, line 23: If you make the context optional in the constructor we don't have to do a major version.

src/index.ts
```diff
@@ -106,7 +120,7 @@ export class UnleashClient extends TinyEmitter {
         }
     }
 
-    public updateContext(context: IContext) {
+    public updateContext(context: IMutableContext) {
```
frikkf, line 123: If this results in a breaking change, we could add another method which does this but update the old method to do the same thing with less strict types.

src/index.ts
mattmurph9: Ya for this commit, you can't make context optional since this package requires appName. appName is part of the Unleash Context. I still think this commit (e16704a) is the most in-line with Unleash Context, but the newer commits are also pretty clean and don't introduce a change to the api.

src/index.ts
mattmurph9: 8cb9aa9 This is now a non-breaking change, while maintaining the advantages of the previous commits. I still think the previous breaking change is more in line with Unleash Context, but in the interest of speed and no breaking changes, I think this is best

src/index.test.ts
mattmurph9: Good question. I also thought the same thing, but it comes back null when it's not in the search params

src/index.ts
ivarconr: I agree with your latest change. Lets try not to do a breaking change for now. We can always do that later.

src/index.ts
ivarconr: I think it is fine, but maybe we should add a console warning if the context object contains either `appName` og `environment`? The typescript definition only helps if the developer uses the type definition.

ivarconr:
Thanks for doing this @mattmurph9. I need to verify that the current `unleash-proxy` actually support the properties params. I think it will, but want to verify, have some time later today :100: 

Also, because of the change in `updateContext` I want this to be a minor release, and that we document the changes. It will only be a breaking for users using the type definitions, and they will get a warning in their IDE.

src/index.test.ts
```diff
@@ -172,9 +172,23 @@ test('Should include context fields on request', async () => {
         [JSON.stringify(data), { status: 200 }],
         [JSON.stringify(data), { status: 304 }],
     );
-    const config: IConfig = { url: 'http://localhost/test', clientKey: '12', appName: 'web', environment: 'prod' };
+    const context: IMutableContext = {
+        userId: '123',
+        sessionId: '456',
+        remoteAddress: 'address',
```
frikkf, line 178: Can we add hostName to the context?

ivarconr: Would that be the hostName of the end-user? Or the service hostName?

src/index.ts
mattmurph9: Something like this @ivarconr? ed1ab8b But would the user even get this far? I would hope even if a user is consuming the package with js, ts would still catch that context is not the right shape and throw an error.

ivarconr: yes. Looking at the code, this is not changed from how it was before, but still nice to help the developers.

ivarconr:
> would yield url search params properties.property1=property1Value&properties.property2=property2Value

It would be beneficial if it could format the properties on as query params a bit different:
`properties[property1]=property1Value&properties[property2]=property2Value`

using `obj[prop]` it looks like nested objects will be supported out of the box with express and other web frameworks. While `object.prop` would require custom parsing in express (current proxy implementation).

src/index.test.ts
frikkf: The hostName of the end-user.
The use case is to send "window.location.hostname" as the domain of the used webapp as a parameter. This can enable staging environments and dev environments to have experimental features toggled

I see that remoteAddress can be used for this. 

I do not know what the unleash community thinks about this. As it seems it has not been thought of.

mattmurph9:
> > would yield url search params properties.property1=property1Value&properties.property2=property2Value
> 
> It would be beneficial if it could format the properties on as query params a bit different:
> `properties[property1]=property1Value&properties[property2]=property2Value`
> 
> using `obj[prop]` it looks like nested objects will be supported out of the box with express and other web frameworks. While `object.prop` would require custom parsing in express (current proxy implementation).

@ivarconr sounds good! I'm not familiar with how this would be supported out of the box, but the approach sounds viable to me. What should nested objects look like in the url search params?

mattmurph9:
> It would be beneficial if it could format the properties on as query params a bit different:
> `properties[property1]=property1Value&properties[property2]=property2Value`

Fixed in 84f976a

ivarconr:
> Do we want to allow nested properties?

No SDKs supports this today, so I would rather not add this now, as it is unclear how the proxy should handle this.

rbtcollins:
Indeed, there cannot be nested objects, as properties is a str:str hash.

mattmurph9:
@ivarconr is there anything left blocking merging this PR? Let me know if you want any other changes!

ivarconr:
thanks for your work. I general I agree with the approach. I will try to get through this tonight or tomorrow to make sure there are no extra details we need to fix.

mattmurph9:
No problem! I look forward to seeing it merged 🚀

ivarconr (approved):
Looks good, and even discovered a small regression in the current unleash-proxy on how it deals with properties while testing this one! :+1:

ivarconr:
released as part of `1.1.0`

PR #21, opened 2021-05-25
Added support for react-native mobile apps

vedyed:
Added react-native async-storage support for setting and getting items. The current storage provider uses `window.localStorage` which is applicable for websites and doesn't work for mobile apps. 

Tested and verified.

### DIFF
--- a/examples/react-app/src/App.js
+++ b/examples/react-app/src/App.js
@@ -9,7 +9,7 @@ function App({unleash}) {
     unleash.on('update', () => {
       setData({ demoToggle: unleash.isEnabled('proxy.demo') });
     })
-  });
+  }, []);
 
 
   return (


src/storage-provider-local.ts
```diff
@@ -7,6 +8,14 @@ export default class LocalStorageProvider implements IStorageProvider {
...
             const key = `${this.prefix}:${name}`;
+            try {
+                AsyncStorage.setItem(key, repo).then((e: any) => {
+                    console.log(`Unleash local save completed for ${key} with value ${repo} and response is ${JSON.stringify(e)}`);
+                });
+            } catch(ex) {
+                console.error(ex);  
+            }
+            // below line is not needed for react-native mobile app
             window.localStorage.setItem(key, repo);
```
ivarconr, line 19: Should we just remove the `winow.localStorage` call all-together? I guess AsyncStorage is used on web as well?

vedyed: Retained this line so that the same package can be used for both web and mobile apps.

vedyed: We can remove this if you're goal is to have separate packages for web and mobile

vedyed: I believe Async-storage cannot be used for websites

vedyed: Suggestion: 
Create a new repo `unleash-proxy-client-react-native` from current `unleash-proxy-client-js` and then we can merge this PR to the newly created repo(we can remove window.localStorage as it is exclusive for react-native mobile apps)

ivarconr:
I think Async Storage can be used on web as well. Need to validate.

https://github.com/react-native-async-storage/async-storage/releases/tag/v1.9.0

PR #22, opened 2021-06-04
use AsyncStorage to make it usable for react native as well. 

ivarconr:


ivarconr:
I have done a few changes to your code @vedyed. AsyncStorage uses local-storage behind the scenes on web so that part just works. Can you confirm that this SDK now will be usable on react-native as well?

PR #30, opened 2021-09-07
Feat / nodejs support

anttiviljami:
Use `global` instead of `window` and allow passing a `fetch` instance.

Example nodejs usage:

```typescript
import { UnleashClient } from 'unleash-proxy-client';
import fetch from 'node-fetch';

const store = new Map();
const inMemoryStorage: IStorageProvider = {
  save: async (key: string, value: any) => {
    store.set(key, value);
  },
  get: async (key: string) => store.get(key),
};

const unleash = new UnleashClient({
  url: config.UNLEASH_PROXY_URL,
  clientKey: config.UNLEASH_PROXY_SECRET,
  appName: config.UNLEASH_APP_NAME,
  storageProvider: inMemoryStorage,
  fetch,
});

unleash.start();
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import { TinyEmitter } from 'tiny-emitter';
 import Metrics from './metrics';
 import IStorageProvider from './storage-provider';
 import LocalStorageProvider from './storage-provider-local';
+import InMemoryStorageProvider from './storage-provider-inmemory';
 
 export interface IStaticContext {
     appName: string;
@@ -18,7 +19,6 @@ export interface IMutableContext {
 }
 
 export type IContext = IStaticContext & IMutableContext;
-export { IStorageProvider }
 
 export interface IConfig extends IStaticContext {
     url: string;
@@ -240,3 +240,6 @@ export class UnleashClient extends TinyEmitter {
         }
     }
 }
+
+// export storage providers from root module
+export { IStorageProvider, LocalStorageProvider, InMemoryStorageProvider }
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { TinyEmitter } from 'tiny-emitter';
 import Metrics from './metrics';
-import IStorageProvider from './storage-provider';
+import type IStorageProvider from './storage-provider';
 import LocalStorageProvider from './storage-provider-local';
 import InMemoryStorageProvider from './storage-provider-inmemory';
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,7 +64,7 @@ export class UnleashClient extends TinyEmitter {
     private etag: string = '';
     private metrics: Metrics;
     private ready: Promise<void>;
-    private fetch?: typeof global['fetch'];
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -100,6 +103,7 @@ export class UnleashClient extends TinyEmitter {
             } 
             resolve();    
         });
+        this.fetch = fetch ?? global.fetch
```
ivarconr, line 106: Could you please explain why will `global.fetch` work in a browser?

ivarconr:
Thanks for providing this. Its good to have the ability to use the proxy also from node.js. 

I would like to highlight that this SDK is meant to operate in a "defined unleash context". 

It will (often) not make sense to use this SDK in a "serverless function", if you need to change the context for every request. In that scenario it would probably be better to call the unleash proxy over http for every call, and maybe utilize a http library with caching built in for similar queries.

src/index.ts
anttiviljami: Looks like `global` is only supported in Chromium. The correct way would be to use `globalThis`, which is widely supported. For IE support, we should fall back to `window`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

ivarconr: thanks. this aligns with my understanding, was thinking it could be some magic in the build step i was unaware of. Had too look up globalThis, and found a nice writeup for others https://blog.logrocket.com/what-is-globalthis-why-use-it/

ivarconr:
I was thinking we could move the inMemory store inside the library as well:

```js
import { UnleashClient, InMemoryStorage } from 'unleash-proxy-client';
import fetch from 'node-fetch';

const unleash = new UnleashClient({
  url: config.UNLEASH_PROXY_URL,
  clientKey: config.UNLEASH_PROXY_SECRET,
  appName: config.UNLEASH_APP_NAME,
  storageProvider: new InMemoryStorage(),
  fetch,
});

unleash.start();
```

wdyt @anttiviljami ?

anttiviljami:
Sure, makes sense. Makes it easier to adopt. Just let me know if you’d like me to make that change

ivarconr:
> Sure, makes sense. Makes it easier to adopt. Just let me know if you’d like me to make that change

Yes please 👍😎

anttiviljami:
@ivarconr added `InMemoryStorageProvidider` with https://github.com/Unleash/unleash-proxy-client-js/pull/30/commits/c83317eb258920512c972eb156ffed8f544eae96 👍

anttiviljami:
Hi @ivarconr! Anything I can help with to get this merged? 😎

ivarconr:
Hi, sorry, last week was quite full. 

Looking in to it now! 👍🏼

ivarconr (approved):

ivarconr:
I found some minor issues in testing before the release. Sorting that out hopefully tomorrow.

ivarconr:
I think it is all sorted out with this one. 
https://github.com/Unleash/unleash-proxy-client-js/commit/c040f14d48f4740fd459c2d4c0fe7956da35f678

PR #38, opened 2021-09-27
adding boostrap feature

ykhedher:


ykhedher:
@ivarconr  Shall I keep the code style or restore as as it was before ?

package.json
```diff
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@react-native-async-storage/async-storage": "^1.15.5",
-    "tiny-emitter": "^2.1.0"
+    "tiny-emitter": "^2.1.0",
+    "unleash-proxy-client": "^1.5.0"
```
ivarconr, line 47: The proxy client should not depend on itself.

src/index.ts
```diff
@@ -1,10 +1,11 @@
...
-import Metrics from './metrics';
-import type IStorageProvider from './storage-provider';
-import LocalStorageProvider from './storage-provider-local';
-import InMemoryStorageProvider from './storage-provider-inmemory';
+import { TinyEmitter } from "tiny-emitter";
+import Metrics from "./metrics";
+import { boostrapData } from "./boostrapData";
+import type IStorageProvider from "./storage-provider";
+import LocalStorageProvider from "./storage-provider-local";
+import InMemoryStorageProvider from "./storage-provider-inmemory";
```
ivarconr, line 6: Please use single quote

src/index.ts
```diff
@@ -1,10 +1,11 @@
...
 
-const DEFINED_FIELDS = ['userId', 'sessionId', 'remoteAddress'];
+const DEFINED_FIELDS = ["userId", "sessionId", "remoteAddress"];
```
ivarconr, line 8: Please use single quote

src/index.ts
```diff
@@ -16,258 +17,285 @@ export interface IMutableContext {
     sessionId?: string;
     remoteAddress?: string;
     properties?: {
-        [key:string]: string,
+        [key: string]: string;
```
ivarconr, line 20: This formatting is breaking or general rules.

ivarconr:
Its a bit hard to review when the formatting is this off. I think we need to fix that first!

ivarconr:
> @ivarconr Shall I keep the code style or restore as as it was before ?

yes please!

package.json
ykhedher: I think it was added when i run `npm install`

src/index.ts
ykhedher: That's the prettier effect. I'll change it.

ykhedher:
Check [NEW PR (new branch)](https://github.com/Unleash/unleash-proxy-client-js/pulls)

PR #49, opened 2021-11-19
Feat/send events

FredrikOseberg:
Adds capabilities to listen on events when doing the isEnabled and getVariant calls. These are emitted through the event listener, and can be picked up by the client and sent to any analytics provider.


src/index.ts
```diff
@@ -170,30 +191,43 @@ export class UnleashClient extends TinyEmitter {
         }
     }
 
+    public createCustomEvent(toggleName: string, action: string) {
+        return this.eventsHandler.createCustomEvent(
+            this.context,
+            toggleName,
```
ivarconr, line 197: toggleName needs to be optional.

ivarconr:
Looks good. I think we can work with this. We should just create an alpha release directly from this branch.

src/index.ts
```diff
@@ -170,30 +191,43 @@ export class UnleashClient extends TinyEmitter {
         }
     }
 
+    public createCustomEvent(action: string, toggleName?: string) {
+        return this.eventsHandler.createCustomEvent(
+            this.context,
+            action,
+            toggleName
+        );
```
ivarconr, line 199: this needs to be emitted?

src/index.ts
```diff
@@ -191,12 +192,13 @@ export class UnleashClient extends TinyEmitter {
         }
     }
 
-    public createCustomEvent(action: string, toggleName?: string) {
-        return this.eventsHandler.createCustomEvent(
+    public sendCustomEvent(toggleName: string) {
+        const event = this.eventsHandler.createCustomEvent(
             this.context,
-            action,
```
FredrikOseberg: We should keep the action.

src/index.ts
```diff
@@ -191,12 +192,13 @@ export class UnleashClient extends TinyEmitter {
...
             toggleName
         );
+        this.eventsHandler.addEvent(event);
```
FredrikOseberg: We can probably also remove this for now.

chriswk (commented):
Why the sudden change to double quotes in src/index.ts ?

FredrikOseberg:
> Why the sudden change to double quotes in src/index.ts ?

There is a war between different editors going on right now.

PR #50, opened 2021-11-30
task: Allow turning off proxy refresh

chriswk:
After a meeting with customer that wanted the possibility to disable
refreshing, they only needed toggle status updated at startup.

This PR adds the possibility to not fetch via a new `disableRefresh` flag, or by setting refreshInterval to `0` the client will no longer perform refreshes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,19 +89,20 @@ if(variant.name === 'blue') {
 
 The Unleash SDK takes the following options:
 
-| option   | required | default | description |
-|----------|----------|---------|-------------|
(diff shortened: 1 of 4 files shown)


ykhedher (approved):
LGTM 👍.

ivarconr:
Nice, I would love to see the available option also documented in the available options section in the README.md:
https://github.com/Unleash/unleash-proxy-client-js#available-options

chriswk:
Will update the documentation before merging

ivarconr (approved):

PR #54, opened 2021-12-09
Fix/metrics

FredrikOseberg:


src/metrics.ts
```diff
@@ -103,51 +122,27 @@ export default class Metrics {
         }
     }
 
-    private startTimer(timeout?: number) {
+    private startTimer(): void {
         if (this.disabled) {
-            return false;
+            return;
```
ivarconr, line 127: Can remove this guard. startTimer will never happen if metrics are disabled, protected inside start().

ivarconr (approved):

FredrikOseberg:
[Fix: metrics in unleash-proxy-client-js](https://trello.com/c/kaC6deTW/656-fix-metrics-in-unleash-proxy-client-js)

PR #60, opened 2022-01-13
fix: do not include context fields with null value

ivarconr:
Currently we are looping over the context fields and add all of them with their respective value to as URL parameters. `searchParams.append(..)` will gladly convert `null` and `undefined` and convert these into string values, which is kind of surprising. 

```js
const urlWithQuery = new URL("https://getunleash.io");

// Using null
urlWithQuery.searchParams.append('a', null);
urlWithQuery.searchParams.has('a');
> true
urlWithQuery.searchParams.get('a');
> "null"

// using undefined
urlWithQuery.searchParams.append('b', undefined);
urlWithQuery.searchParams.has('b');
> true
urlWithQuery.searchParams.get('b');
> "undefined"
```

this PR adds filtering of context fields which has value `null` or `undefined`.


FredrikOseberg (approved):
LGTM

PR #63, opened 2022-01-16
Add a default value parameter to isEnabled

snowfrogdev:
Change isEnable to comply with Unleash's [official Client Specification](https://docs.getunleash.io/client-specification#the-basics). It stipulates that
> The basic isEnabled method should also accept a default value. This should be used if the client does not know anything about a particular toggle. If the user does not specify a default value, false should be returned for unknown feature toggles.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -163,9 +163,9 @@ export class UnleashClient extends TinyEmitter {
         return [...this.toggles];
     }
 
-    public isEnabled(toggleName: string): boolean {
+    public isEnabled(toggleName: string, defaultValue = false): boolean {
         const toggle = this.toggles.find((t) => t.name === toggleName);
-        const enabled = toggle ? toggle.enabled : false;
+        const enabled = toggle ? toggle.enabled : defaultValue;
         this.metrics.count(toggleName, enabled);
         return enabled;
     }


thomasheartman:
Thanks for the PR and for your time! 😄 However, there is an issue with doing this for the front-end clients, so the lack of this feature is actually intentional. I'll explain below, but if you have a use case where this would somehow help you out, I'd be happy to hear it ☺️  

## Why front-end clients don't have fallbacks

Front-end clients fetch their features via the Unleash Proxy instead of directly from the Unleash instance itself. For privacy reasons, as outlined in [the Unleash Proxy documentation](https://docs.getunleash.io/sdks/unleash-proxy), the Proxy does not provide the client with the names of toggles that are disabled for the user.

This means that the client cannot distinguish between disabled toggles and missing toggles. Because the purpose of the fallback values is to provide a value if the toggle is _missing_ and not if it is _disabled_, we have decided to not implement this for the front-end clients.

For the time being, there are no plans for implementing this functionality for the front-end clients. However, if we do change the proxy to also provide names of disabled toggles (via a config option, for instance), then this would make sense.

snowfrogdev:
> Thanks for the PR and for your time! 😄 However, there is an issue with doing this for the front-end clients, so the lack of this feature is actually intentional. I'll explain below, but if you have a use case where this would somehow help you out, I'd be happy to hear it ☺️
> 
> ## Why front-end clients don't have fallbacks
> Front-end clients fetch their features via the Unleash Proxy instead of directly from the Unleash instance itself. For privacy reasons, as outlined in [the Unleash Proxy documentation](https://docs.getunleash.io/sdks/unleash-proxy), the Proxy does not provide the client with the names of toggles that are disabled for the user.
> 
> This means that the client cannot distinguish between disabled toggles and missing toggles. Because the purpose of the fallback values is to provide a value if the toggle is _missing_ and not if it is _disabled_, we have decided to not implement this for the front-end clients.
> 
> For the time being, there are no plans for implementing this functionality for the front-end clients. However, if we do change the proxy to also provide names of disabled toggles (via a config option, for instance), then this would make sense.

Makes total sense. I figured it was probably intentional but I couldn't think of why. Thanks for the explanation.

Meemaw:
What about if the unleash proxy is down, and the client can't properly initialise? In that case, we would like our feature flags to have sensible fallbacks.

thomasheartman:
@Meemaw Absolutely! Have you checked out the [bootstrap options](https://github.com/Unleash/unleash-proxy-client-js#bootstrap)? You can use that property to pass the exact configuration you want at startup. It'll use that configuration until it gets a newer one from the proxy. Does that answer your question? ☺️

PR #69, opened 2022-02-03
fix: upgrade jest-fetch-mock

ivarconr:


ivarconr:
Fails because of typescript. All test runs green if we disable typescript checking in the tests.

kei95:
Hi @ivarconr! I open a new PR on top of this branch to fix these typescript failures in tests. That would be great if you could have a look when you get a chance.
https://github.com/Unleash/unleash-proxy-client-js/pull/88

PR #74, opened 2022-03-17
docs: Add a section for the stop() method

ivarconr:


README.md
```diff
@@ -131,6 +131,13 @@ unleash.on('update', () => {
 
 You may provide a custom session id via the "context". If you do not provide a sessionId this SDK will create a random session id, which will also be stored in the provided storage (local storage). By always having a consistent sessionId available ensures that even "anonymous" users will get a consistent experience when feature toggles is evaluated, in combination with a gradual (percentage based) rollout. 
 
+### Stop the SDK
+Sometimes you would like to stop the Unleash SDK. The effect of stopping the SDK is that it will not check for any more updates and not send metrics to the Unleash proxy anymore.
```
thomasheartman, line 135: Great! I've reworded it silghtly, but it's up to you whether we change it or not ☺️ 

```suggestion
You can stop the Unleash client by calling the `stop` method. Once the client has been stopped, it will no longer check for updates or send metrics to the server.
```

README.md
```diff
@@ -131,6 +131,13 @@ unleash.on('update', () => {
...
+
```
thomasheartman, line 136: Can you restart a client after stopping it? I think we should mention that explicitly. This suggestion says you can, but we may need to change it to "can not".

```suggestion

A stopped client _can_ be restarted.

```

thomasheartman (changes requested):
Super! 🎉  Got one rewording suggestion and one question that I'd like an answer to and some more info on; all mentioned inline.

thomasheartman (approved):
Love it! ❤️

PR #76, opened 2022-04-04
BREAKING CHANGE: remove built-in support for react native async storage

Dzixxx:
[Breaking Change] v2

Removing built-in support for react-native (will become optional as custom storageProvider provided by developer itself)
* if will remove a lot of installed dependencies (for react-native it's huge...)
* it will reduce amount of packages that needs to be maintained (dependabot, renovate...)
* it's a suggestion to create own client for react-native if more custom things are needed 
* instead of `react-native-async-storage` which fallbacks to window.localStorage I just used window.localStorage

Kind regards
Dziczek!

### DIFF
--- a/src/storage-provider-local.ts
+++ b/src/storage-provider-local.ts
@@ -1,4 +1,3 @@
-import AsyncStorage from '@react-native-async-storage/async-storage';
 import type IStorageProvider from './storage-provider';
 
 export default class LocalStorageProvider implements IStorageProvider {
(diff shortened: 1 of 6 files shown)


src/storage-provider-local.ts
```diff
@@ -8,16 +7,16 @@ export default class LocalStorageProvider implements IStorageProvider {
...
         try {
-            await AsyncStorage.setItem(key, repo);
+            window.localStorage.setItem(key, repo);
         } catch (ex) {
             console.error(ex);
         }
     }
 
-    public async get(name: string) {
+    public get(name: string) {
```
ivarconr, line 16: Even though we do not need `async` marker on this method i think I would prefer to have it, to make sure we remember to await the get method also, sense this will be the default implementation. I can easily see myself (or someone else) forgetting the "await" in the calling code, if this method is not marked as async in the default implementation. This will lead to bugs, only visible when using a custom store.

Dzixxx: Evening! Not problem to get rid of async to but U've defined interface which is Promise based :) 
```typescript
// src/storage-provider.ts
export default interface IStorageProvider {
    save: (name: string, data: any) => Promise<void>;
    get: (name: string) => Promise<any>;
}
```

Dzixxx:
(rebasing ☝️)

Dzixxx:
@ivarconr ping 😅

src/storage-provider-local.ts
ivarconr: good point :D

ivarconr (approved):

PR #99, opened 2022-07-27
fix: build paths

FredrikOseberg:
* Moves tests into src so that we don't change the folder structure of the build output


sighphyre (approved):
You can get a hell yeah

sighphyre (changes requested):
...but also, let's get the tests working

sighphyre (approved):
LGTM

PR #102, opened 2022-08-10
feat: add script to check if release is ok

FredrikOseberg:
This PR adds a script that will verify that the package.json main field matches the build output. This is an action item from our post-mortem following a bug where typescript would change the output of the build folder when importing files outside of src.


scripts/check-release.sh
```diff
@@ -0,0 +1,11 @@
+
+# Get the main field from package.json
+FILE=$(cat ./package.json | grep main | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
```
gastonfournier, line 3: This can be slightly simplified:
```suggestion
FILE=$(grep '"main"' package.json | sed 's/.*:\s*"\(.*\)".*/\1/')
```

There's an alternative which will also validate if the json is valid, but only works with python installed (in my case, the command is `python3`): 
```console
cat package.json | python -c 'import json,sys;obj=json.load(sys.stdin);print(obj["main"])';
```

gastonfournier (approved):
The command is valid, just leaving you a suggestion

scripts/check-release.sh
FredrikOseberg: Thanks! I reverted back because this command: `echo $(grep '"main"' package.json | sed 's/.*:\s*"\(.*\)".*/\1/')` yields this output for me: `"main": "./build/index.js",`

And I just need the path for the rest of the command to work

gastonfournier: Strange... I knew sed had different behaviors in different systems. It must be some kind of nuance I'd need to check.

In any case:
```suggestion
FILE=$(grep '"main":' ./package.json | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
```
should work fine

Issue #59, opened 2022-01-06
Allow to pass relative url as config


wodka:
Currently when passing something like '/unleash' into the config an error is thrown.

```ts
new UnleashClient({
  url: '/unleash',
})
```

Throws:
```
TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:371:5)
    at onParseError (node:internal/url:552:9)
    at new URL (node:internal/url:628:5)
    at new UnleashClient (node_modules/unleash-proxy-client/build/index.js:121:21)
```

https://github.com/Unleash/unleash-proxy-client-js/blob/main/src/index.ts#L123

FredrikOseberg:
Thanks for this suggestion, I'll bring it up with the team.

olav:
@wodka 

The upcoming support for `URL` instances in config objects can help when using paths, though a `base` is still required:

```ts
new UnleashClient({
  url: new URL('/unleash', document.baseURI),
})
```

We could perhaps do this automatically for relative URL strings, but not sure if necessary.

PR #103, opened 2022-08-10
feat: allow using URL instances in config objects

olav:
Resolves https://github.com/Unleash/unleash-proxy-client-js/pull/72.

Updates https://github.com/Unleash/unleash-proxy-client-js/issues/59.


andreas-unleash (approved):
LGTM

PR #107, opened 2022-09-07
Add prerelease argument to create-release

chriswk:
This tags alphas and betas as prerelease so our Latest version does not display beta's as latest versions


thomasheartman (approved):
Oh, this looks neat! Do we do any other prereleases? Like `rc` or something? Anyway, if we don't, then I think this is good to go!

PR #109, opened 2022-09-13
enhancement: emit errors if response is not ok

### LINKED ISSUE #108: Expand error reporting functionality for Http errors
### Describe the feature request

I'd like the unleash-proxy client to emit error events on http errors in the 400-500 range.

### Background

Currently, unleash-proxy-client depends on the fetch implementation to emit error events, and we've now had a couple of cases where users expected to be able to react errors from the client also for http errors in the 4-500 range. 

### Solution suggestions

On refreshing the toggles, the proxy client should check the status of the response, if it is in the 4-500 range it should emit an Error event allowing consumers of the unleash events to further react to what went wrong

Further improvement should also probably take into consideration that a 401 on initialization should lead to stopping poll.

chriswk:
This now checks if response is not ok, if it is not, we emit an error object with a type and a statusCode.

I'm wondering if we should add more logic when we receive a 401 and just stop the poll; but that can be a different improvement.
fixes #108

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ unleash.on('update', () => {
 
 #### Available events:
 
-- **error** - emitted when an error occurs on init, or when fetch function fails. The error object is sent as payload.
+- **error** - emitted when an error occurs on init, or when fetch function fails, or when fetch receives a non-ok response object. The error object is sent as payload.
 - **initialized** - emitted after the SDK has read local cached data in the storageProvider. 
 - **ready** - emitted after the SDK has successfully started and performed the initial fetch towards the Unleash Proxy. 
 - **update** - emitted every time the Unleash Proxy return a new feature toggle configuration. The SDK will emit this event as part of the initial fetch from the SDK.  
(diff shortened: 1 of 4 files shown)


chriswk:
I'll add a couple of tests verifying this behaviour.

ivarconr:
> I'm wondering if we should add more logic when we receive a 401 and just stop the poll; but that can be a different improvement.

Jeah, I guess the user can also call stop themselves in that scenario if they like. A 401 could be temporary (e.g. incorrectly sending traffic to a different server)

ivarconr (approved):

fivaz:
how can I omit those errors ? my console log is being spammed by the same error (my feature flag is working fine)

grovertb:
Hi everyone, just like @fivaz it shows me the same error. Your support please on how to bypass it, my feature flag works fine.

melloware:
I think you intrdouced this issue: https://github.com/Unleash/unleash-proxy-client-js/issues/111

I think you are missing 304 in your test its reporting as an error when its an OK.

melloware:
I think this line...

```js
if (response.ok && response.status !== 304) {
```

needs to be

```js
if (response.ok || response.status === 304) {
```

PR #115, opened 2022-09-23
Fix/revert browser field

FredrikOseberg:
* Do not emit error for 304s
* Revert browser field in package.json


sighphyre (approved):
Thank you!

PR #120, opened 2022-10-10
Feat/impression data

ivarconr:
Make it possible to enable impression data for all isEnabled / getVariant calls. This PR also introduces a "impressionData" flag to the event itself, so the implementer can choose to discard events when we do not know whether impressionData is enabled for this toggle. 

The reason we need this override in the proxy-sdk is that it simply does not know anything about disabled toggles (by design, to not leak information to the end-user about toggles). This makes it impossible to get impression data for disabled toggles, as they are not exposed to the frontend at all. 

Todo:

- [ ] Add documentation for the new config property
- [ ] Add unit tests for`impressionDataAll`


ivarconr:
the changeset is way off. I must have been using an old main branch. I will do some cleanup .

PR #121, opened 2022-10-10
feat: add overide to get all impression events

ivarconr:
Make it possible to enable impression data for all isEnabled / getVariant calls. This PR also introduces a "impressionData" flag to the event itself, so the implementer can choose to discard events when we do not know whether impressionData is enabled for this toggle.

The reason we need this override in the proxy-sdk is that it simply does not know anything about disabled toggles (by design, to not leak information to the end-user about toggles). This makes it impossible to get impression data for disabled toggles, as they are not exposed to the frontend at all.

**Todo:**

- [x]  Add documentation for the new config property
- [x]  Add unit tests forimpressionDataAll

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ The Unleash SDK takes the following options:
 | bootstrapOverride | no| `true` | Should the bootstrap automatically override cached data in the local-storage. Will only be used if bootstrap is not an empty array.     | 
 | headerName        | no| `Authorization` | Provides possiblity to specify custom header that is passed to Unleash / Unleash Proxy with the `clientKey` | 
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used. | 
-| impressionDataAll | no| `false` | Allows you to trigger "impression" events for all `getToggle` and `getVariant` invocations. This is particular useful for "disabled" feature toggles which are not visible to frontend SDKs. | 
+| impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. | 
 
 ### Listen for updates via the EventEmitter
 
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ The Unleash SDK takes the following options:
 | bootstrapOverride | no| `true` | Should the bootstrap automatically override cached data in the local-storage. Will only be used if bootstrap is not an empty array.     | 
 | headerName        | no| `Authorization` | Provides possiblity to specify custom header that is passed to Unleash / Unleash Proxy with the `clientKey` | 
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used. | 
-
+| impressionDataAll | no| `false` | Allows you to trigger "impression" events for all `getToggle` and `getVariant` invocations. This is particular useful for "disabled" feature toggles which are not visible to frontend SDKs. | 
 
 ### Listen for updates via the EventEmitter
 
--- a/src/events-handler.ts
+++ b/src/events-handler.ts
@@ -10,16 +10,16 @@ class EventsHandler {
         context: IContext,
         enabled: boolean,
         featureName: string,
-        impressionData: boolean,
         eventType: string,
+        impressionData?: boolean,
         variant?: string
     ) {
         const baseEvent = this.createBaseEvent(
             context,
             enabled,
             featureName,
+            eventType,
             impressionData,
-            eventType
         );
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -107,7 +107,7 @@ The Unleash SDK takes the following options:
 | bootstrapOverride | no| `true` | Should the bootstrap automatically override cached data in the local-storage. Will only be used if bootstrap is not an empty array.     | 
 | headerName        | no| `Authorization` | Provides possiblity to specify custom header that is passed to Unleash / Unleash Proxy with the `clientKey` | 
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used. | 
-
+| impressionDataAll | no| `false` | Allows you to trigger "impression" events for all `getToggle` and `getVariant` invocations. This is particular useful for "disabled" feature toggles which are not visible to frontend SDKs. |
```
thomasheartman, line 110:
```suggestion
| impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. | 
```

src/events-handler.ts
```diff
@@ -33,6 +35,7 @@ class EventsHandler {
         context: IContext,
         enabled: boolean,
         featureName: string,
+        impressionData: boolean,
```
thomasheartman, line 38: Would it make sense to use a name like `impressionDataEnabled` instead? It might make the name slightly clearer (seeing as it's a boolean and not _actual_ impression data). Happy to be overruled here, though.

thomasheartman (commented):
I like the direction this is going in and think it makes a lot of sense  😄 

I wonder whether reducing `impressionData` to a boolean is a good idea, though. It's actually one of **three** states, so using a boolean, you lose information:
1. Toggle is enabled and has impressionData enabled
2. Toggle is enabled and does **not** have impression data enabled
3. Toggle is disabled, so we don't know.

Would it be better to model it as `boolean | "unknown"` or something?

src/index.test.ts
```diff
@@ -1114,6 +1114,79 @@ test('Should call getVariant event when impressionData is true', (done) => {
...
+        bootstrap,
+        impressionDataAll: true,
+    };
+    const client = new UnleashClient(config);
+    client.start();
+
+    client.on(EVENTS.READY, () => {
+        const isEnabled = client.getVariant('impression-variant');
+        expect(isEnabled).toBe(true);
+    });
```
thomasheartman, line 1184: There's no test for whether impressionData events are returned when a toggle is `false` but `impressionDataAll` is true: is that supposed to have the same behavior?

src/events-handler.ts
ivarconr: jeah I think it would be better, but I just wanted top keep it the same as the featureToggle property, which currently is "impressionData". 

Wdyt?

src/index.test.ts
ivarconr: yes, I should add a test case for this as well. I did it for getVariant.

ivarconr:
> I like the direction this is going in and think it makes a lot of sense smile
> 
> I wonder whether reducing `impressionData` to a boolean is a good idea, though. It's actually one of **three** states, so using a boolean, you lose information:
> 
> 1. Toggle is enabled and has impressionData enabled
> 2. Toggle is enabled and does **not** have impression data enabled
> 3. Toggle is disabled, so we don't know.
> 
> Would it be better to model it as `boolean | "unknown"` or something?

You make a strong argument here. We could make the `impressionData` property optional and only include it when the toggle definition exists. This would work in js environments anyway as we have the undefined as falsy behavior.

src/events-handler.ts
thomasheartman: I'm open to either. As long as we've considered the option, I'm happy for you to make a decision on it ☺️  I see how keeping it the same as the toggle property makes sense, but I **personally** think it makes it clearer to use `impressionDataEnabled` or even `impressionDataEnabledForFeature`. It _could_ cause some confusion when handling an impression data event that you also have a subproperty called `impressionData`, so I'm leaning towards being more explicit here.

thomasheartman:
@ivarconr Yeah, I think using `boolean | undefined` would work fine too. That does cover the use cases, though it makes it possible to "forget" to add it. When we make it required, you **have** to provide a value, which may prevent hard-to-spot bugs in the future 💁🏼

ivarconr:
> boolean | "unknown"

this will only use if the user of this SDK also uses typescript. The only alternative is to use string values, which feels a bit verbose in this use case. I think the only reason to enable this flag is that you really want impression events for all, and optionally have your own whitelist for what to actually store.

thomasheartman:
Yeah, alright, I can buy that 😄

thomasheartman (approved):
Looks good to me! 🚂  We should make sure to update the Impression data docs for this too 🤓

PR #122, opened 2022-10-26
fix: Include impression data status when not enabled

thomasheartman:
## What

This PR fixes a bug in how impression data is reported when `impressionDataAll` is set to true: When a feature has `impressionData: false` but `impressionDataAll` is set to true, it will now report `impressionData: false` instead of `impressionData: undefined`.

## Why

This is a bug because we should report the state of impression data for a feature when we can. In the original feature PR discussion, it was agreed to make `impressionData` a `boolean | undefined` where it would be reported as a boolean if we knew the state and `undefined` otherwise.

It seems to have been an oversight in the tests.

### What's with the try/catch blocks in the tests?

It's explained more thoroughly in the [callback section of Jest's testing asynchronous code](https://jestjs.io/docs/asynchronous#callbacks), but in short:

If you don't have the try/catch and one of the expectations fail, then the done callback is never called (because the function short circuits). While this does mean that the test fails (this is good), it fails with a message saying that it exceeded the set timeout and not because it failed an expectation (this is bad).

To instead have the correct error message reported, you must catch the error and pass it to the done callback. This will make the test fail and report the correct error message.

Now regarding using `finally`, I did consider that but ended up not doing it for the following reason:

Calling the done callback requires the error for it to fail. This can't be done in a finally block because you don't have access to the error. In theory, you could probably declare an error variable that you set to undefined and then assign in the catch block, but that doesn't feel any cleaner to me.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1186,7 +1186,7 @@ test('Should call isEnabled event when impressionData is false and impressionDat
     client.on(EVENTS.IMPRESSION, (event: any) => {
         expect(event.featureName).toBe('impression');
         expect(event.eventType).toBe('isEnabled');
-        expect(event.impressionData).toBe(undefined);
+        expect(event.impressionData).toBe(false);
         client.stop();
         done();
     });
@@ -1261,7 +1261,7 @@ test('Should call getVariant event when impressionData is false and impressionDa
     client.on(EVENTS.IMPRESSION, (event: any) => {
         expect(event.featureName).toBe('impression-variant');
         expect(event.eventType).toBe('getVariant');
-        expect(event.impressionData).toBe(undefined);
+        expect(event.impressionData).toBe(false);
         client.stop();
         done();
     });
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1023,7 +1023,7 @@ test('Should pass under custom header clientKey', async () => {
     jest.advanceTimersByTime(999);
 });
 
-test('Should call isEnabled event when impressionData is true', (done) => {
+test('Should emit impression events on isEnabled calls when impressionData is true', (done) => {
     const bootstrap = [
         {
             name: 'impression',
@@ -1084,7 +1084,7 @@ test('Should pass custom headers', async () => {
     });
 });
 
-test('Should call getVariant event when impressionData is true', (done) => {
+test('Should emit impression events on getVariant calls when impressionData is true', (done) => {
     const bootstrap = [
         {
             name: 'impression-variant',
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -1184,15 +1184,20 @@ test('Should call isEnabled event when impressionData is false and impressionDat
     });
 
     client.on(EVENTS.IMPRESSION, (event: any) => {
-        expect(event.featureName).toBe('impression');
-        expect(event.eventType).toBe('isEnabled');
-        expect(event.impressionData).toBe(undefined);
-        client.stop();
-        done();
+        try {
```
sighphyre, line 1187: Why the try/catch here? If it's needed, a try/finally might be better here

sighphyre (commented):
This is such a subtle change that I think that if it was worth fixing then it's worth adding a test for. I agree with the source code changes but the new try/catches look odd to me

src/index.test.ts
thomasheartman: Good point! I explained it in the commit, but not in the PR, I see. It's explained more thoroughly in the [callback section of Jest's testing asynchronous code](https://jestjs.io/docs/asynchronous#callbacks), but in short:

If you don't have the try/catch and one of the expectations fail, then the `done` callback is never called (because the function short circuits). While this _does_ mean that the test fails (this is good), it fails with a message saying that it exceeded the set timeout and not because it failed an expectation (this is bad).

To instead have the correct error message reported, you must catch the error and pass it to the `done` callback. This will make the test fail **and** report the correct error message.

Now regarding using finally, I _did_ consider that but ended up not doing it for the following reason:

Calling the `done` callback requires the error for it to fail. This can't be done in a finally block because you don't have access to the error. In theory, you could probably declare an `error` variable that you set to undefined and then assign in the `catch` block, but that doesn't feel any cleaner to me. I'd be happy to take suggestions, though 😄

thomasheartman: A follow-up question, then, could be: why didn't I add try/catch clauses to **all the other tests** that also use the `done` callback? I thought about it, but didn't think it'd be worth it. The tests still fail, even if their error messages are confusing, and the diff would've been much bigger and harder to understand if I _had_ gone and done that. 

However, I could do it now if you think it's sensible? 💁🏼

sighphyre: Right, gosh that is arcane but definitely nice to have!

I think it's fine as is. This is effectively a 4 character change to source code that's already quite a large diff. It's nice to have commits self contained so if you're keen you're welcome to do it in a separate PR but I think this one stands as it is without adding more to it.

sighphyre (approved):
Nice

PR #125, opened 2022-11-29
Feat/post fetching

ivarconr:
In this PR we make it configurable to use POST instead of GET when polling for feature toggle updates. This avoids exposing context fields as query params on the URL, but will reduce the built in cache-ability. 

See also https://github.com/Unleash/unleash-proxy/pull/106

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -56,6 +56,22 @@ test('Should perform an initial fetch as POST', async () => {
     expect(body.context.appName).toBe('webAsPOST');
 });
 
+test('Should perform an initial fetch as GET', async () => {
+    fetchMock.mockResponseOnce(JSON.stringify(data));
+    const config: IConfig = {
+        url: 'http://localhost/test',
+        clientKey: '12',
+        appName: 'webAsGET',
+    };
+    const client = new UnleashClient(config);
+    await client.start();
+    
+    const request = getTypeSafeRequest(fetchMock, 0);
+
+    expect(request.method).toBe('GET');
+});
+
+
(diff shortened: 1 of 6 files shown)


src/util.ts
```diff
@@ -1,2 +1,32 @@
...
+                        )
+                    );
+            } else {
+                urlWithQuery.searchParams.append(
+                    contextKey,
+                    contextValue
+                );
+            }
+        });
+    return urlWithQuery;
```
FredrikOseberg, line 31: Would be nice with a unit test for this.

src/index.test.ts
```diff
@@ -38,6 +38,24 @@ test('Should perform an initial fetch', async () => {
...
+    };
+    const client = new UnleashClient(config);
+    await client.start();
+    
+    const request = getTypeSafeRequest(fetchMock, 0);
+    const body = JSON.parse(request.body as string);
+
+    expect(request.method).toBe('POST');
+    expect(body.context.appName).toBe('webAsPOST');
+});
```
FredrikOseberg, line 57: Should we have the inverse test also? That not setting the property keeps the method as GET.

FredrikOseberg (approved):
Looks good, I've added two suggestions.

src/index.test.ts
ivarconr: yes makes sense.

src/util.ts
ivarconr: Good suggestion. It is implicitly tested, but with be good with a few concrete unit test on this function in isolation.

PR #126, opened 2022-11-29
Feat/store etag

ivarconr:
Store etag in localStorage would make sure that also the initial fetch can be 304 if the user already has the latest toggles in localStorage. Builds on top of [feat/POST-fetching](https://github.com/Unleash/unleash-proxy-client-js/tree/feat/POST-fetching)

I think this is a good idea?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -39,6 +39,7 @@ interface IConfig extends IStaticContext {
     headerName?: string;
     customHeaders?: Record<string, string>;
     impressionDataAll?: boolean;
+    usePOSTrequests?: boolean;
 }
 
 interface IVariant {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -351,10 +356,16 @@ export class UnleashClient extends TinyEmitter {
...
+
                 if (response.ok && response.status !== 304) {
-                    this.etag = response.headers.get('ETag') || '';
+                    const etag = response.headers.get('ETag');
                     const data = await response.json();
-                    await this.storeToggles(data.toggles);
+                    
+                    if(etag && etag !== this.etag) {
+                        await this.storeToggles(data.toggles, etag);
+                    } else if(!etag) {
```
ivarconr, line 366: Needed to support the case where a proxy do not fancy to send 304 responses at all.

PR #130, opened 2022-12-15
Update context waits for start

### LINKED ISSUE #127: updateContext and setContextField don't call fetchToggles until first fetchToggles completes
### Describe the bug

After changing the context via updateContext, a new request for toggles doesn't trigger immediately. This looks like it's because the first fetchToggles and/or ready promise hasn't returned in `start` yet and timerRef isn't set.

Additionally, this means .start can be called multiple times until any chain of promises resolve.

- Start a client.
- Update the context after.
- Observe no request for toggles with the latest context.

### Expected behavior

When calling `updateContext`, a new request for fetchToggles should occur immediately if `start` has already been called.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

Using @unleash/proxy-client-react and we are changing the context in a child component.

### Unleash version

4.17.2

### Subscription type

Pro

### Hosting type

Hosted by Unleash

_No response_

kwasniew:
## About the changes

updateContext wait for start even if we don't await start 

Closes #127

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build/
 node_modules/
 .vscode/
 coverage/
+.idea
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "unleash-proxy-client",
-  "version": "2.4.0-beta.0",
+  "version": "2.4.0",
   "lockfileVersion": 2,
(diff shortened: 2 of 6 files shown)


ivarconr (approved):
Looks good 👍🏼

PR #131, opened 2022-12-22
fix: count variant metrics based on enabled state

FredrikOseberg:
## About the changes

This PR fixes a bug where getVariant would always count as true, disregarding the enabled state of the toggle.

Closes #

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -198,7 +198,7 @@ export class UnleashClient extends TinyEmitter {
                 enabled,
                 toggleName,
                 IMPRESSION_EVENTS.IS_ENABLED,
-                toggle?.impressionData ?? undefined,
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -211,15 +211,15 @@ export class UnleashClient extends TinyEmitter {
         const enabled = toggle?.enabled || false;
         const variant = toggle ? toggle.variant : defaultVariant;
 
-        this.metrics.count(toggleName, true);
+        this.metrics.count(toggleName, enabled);
```
FredrikOseberg, line 214: This line is the important change.

gastonfournier (approved):
:+1: similar to the fix we did for go-client

kwasniew (approved):

PR #134, opened 2023-01-11
fix: nodejs support by default

Tymek:
## About the changes

- [x] Prevent crashing if `window` is undefined without additional configuration
- [x] Fix typo in `InMemoryStorageProvider`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,7 +76,7 @@ const storeKey = 'repo';
 
 export const resolveFetch = () => {
     try {
-        if ('fetch' in window) {
+        if (typeof window !== 'undefined' && 'fetch' in window) {
             return fetch.bind(window);
         } else if ('fetch' in globalThis) {
             return fetch.bind(globalThis);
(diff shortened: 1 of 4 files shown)


src/storage-provider-inmemory.ts
```diff
@@ -1,6 +1,6 @@
 import type IStorageProvider from './storage-provider';
 
-export default class ImMemoryStorageProvider implements IStorageProvider {
+export default class InMemoryStorageProvider implements IStorageProvider {
```
ivarconr, line 3: nice catch, safe to change because it is a default export.

src/index.ts
```diff
@@ -76,7 +76,7 @@ const storeKey = 'repo';
 
 export const resolveFetch = () => {
     try {
-        if ('fetch' in window) {
+        if (typeof window !== 'undefined' && 'fetch' in window) {
```
ivarconr, line 79: why not `if(window && 'fetch' in window)`?

ivarconr (approved):

src/index.ts
Tymek: Checking if `window` is truthy can still trigger TS error "window is not defined"

PR #141, opened 2023-02-09
fix: bootstrap without waiting for promises

Tymek:
## About the changes
Toggles should be instantly ready (synchronous) when bootstrapping. This is needed for SSR with React SDK.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -71,7 +71,6 @@ test('Should perform an initial fetch as GET', async () => {
     expect(request.method).toBe('GET');
 });
 
-
 test('Should have correct variant', async () => {
     fetchMock.mockResponseOnce(JSON.stringify(data));
     const config: IConfig = {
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -89,6 +89,7 @@ export const resolveFetch = () => {
 };
 
 export class UnleashClient extends TinyEmitter {
+    bootstrapped = false;
```
thomasheartman, line 92: Is this essentially a combination of other fields? What does it mean if `bootstrapped` is `true`, but `bootstrap` and `bootstrapOverride` are false? Would it make more sense as a method?

And what's the property used for? I see it shows up in the test, but is it used for anything else?

src/index.test.ts
```diff
@@ -1407,3 +1409,28 @@ test('Should be able to configure UnleashClient with a URL instance', () => {
     const client = new UnleashClient(config);
     expect(client).toHaveProperty('url', url);
 });
+
+test('Should be immediately ready when bootstrapping', () => {
```
kwasniew, line 1413: this test is passing for me even before your changes so I'm wondering what is this change about?

src/index.ts
kwasniew, line 92: I don't think bootstrapped should be part of our public API of the SDK.

kwasniew (changes requested):
The test that shows the new behavior is also passing for me in the old version of the SDK so something is off here.

Issue #146, opened 2023-03-07
`undefined` userId considered in context


SimeonKostov:
### Describe the bug

In our React app, we had been updating the Unleash context's `userId` field in the following way:
```
useEffect(() => {
  updateContext({userId});
}, [userId])
```
until we noticed that there was an issue with the variant splitting for one of our feature flags - a flag with `control` and `challenger` variants, each with 50% weight.

Initially, the **userId** is `undefined` until an account is created, and seems like the `undefined` value is sent to the context as a string because the splitting issue is just for a single flag used before an account is created.

### Steps to reproduce the bug

1. Create a feature flag with 2 variants
2. Assign 50% weight to each variant
3. set `undefined` to the `userId` context field
4. monitor how many times both variants have been assigned

### Expected behavior

When a `userId` is undefined in the context, a `sessionId` should be used in order to distinguish the different users calling the proxy/server.

### Unleash version

unleashorg/unleash-proxy:v0.11.7

### Subscription type

Pro

### Hosting type

Hosted by Unleash

### SDK information (language and version)

@unleash/proxy-client-react v3.5.1, unleash-proxy-client v2.4.2

thomasheartman:
Hi, @SimeonKostov! Thanks for reporting this. But I'm not sure I understand exactly what the bug you are describing is. Could you elaborate?

You say that "there was an issue with the variant splitting for one of our feature flags". What was that issue? You later refer to it as the "splitting issue". What exactly is happening? Are you not seeing the split? Or is it "flip-flopping" between rerenders?  

And in terms of the variant configuration, what have you set as the stickiness? Are you using the "default" stickiness?

Looking at it locally and some of the tests we've got (plus a few more I just added), things seem to be working as expected on my end 🤔  Looking forward to get some clarification on your end.

thomasheartman:
Someone informed me that this probably comes from this [community slack thread](https://unleash-community.slack.com/archives/C03GZBSJV9A/p1677688981623749) that has a lot more context, so I'm putting the link here for that case. 

I don't know exactly what's happening here, but I tried reproducing the case you were seeing in the tests in #147, but it seems to work as expected. 

Regarding `undefined` being stringified to `"undefined"`, shouldn't you see it in the query parameters you send in that case? I also can't reproduce that in the tests in any way. Not ruling it out completely, but I'm not sure why it would behave that way outside the tests 🤔

SimeonKostov:
[...]

PR #147, opened 2023-03-09
fix: Update tests

thomasheartman:
This PR does three things:

- Adds tests for undefined/null properties 
- Updates test titles
- Formats the test suites

The new tests were brought on by #146

## New tests

I've added two (three) new tests.
1. `src/util.test.ts`: Verify that object properties set to `null` or `undefined` are excluded from query params
2. `src/index.test.ts`: Verify that setting a context field to `null` or `undefined` removes it from the context

## Test title updates

Two of the tests in `src/util.test.ts` seems to have had copy-paste mistakes in their titles. Looking at them, I believe the "not" in their descriptions should be removed.

## Formatting

This repo has prettier listed in its deps, but the test files were pretty unformatted. My editor auto-formats on save, so this led to a lot of changes. However, I've isolated the formatting changes into separate commits, so we can easily cut those out if we don't want them.

Diffing with **hidden whitespace** is encouraged.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -71,7 +71,6 @@ test('Should perform an initial fetch as GET', async () => {
     expect(request.method).toBe('GET');
 });
 
-
 test('Should have correct variant', async () => {
     fetchMock.mockResponseOnce(JSON.stringify(data));
     const config: IConfig = {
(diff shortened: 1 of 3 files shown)


andreas-unleash (approved):
LGTM

PR #148, opened 2023-03-16
fix: use custom headers for metrics requests too

### LINKED ISSUE #142: `customHeaders` aren't included in metrics requests
### Describe the bug

According to [this slack thread](https://unleash-community.slack.com/archives/CGP2MCHPF/p1676061868954369) and @kwasniew's digging:

Custom headers aren't included in metrics requests.

1. Initialize the proxy client with custom headers
2. Check the network tab and watch for metrics requests
3. Confirm that the metrics requests **don't** contain the custom headers

### Expected behavior

When I provide custom headers to the client, I'd expect it to use these headers for all communication with the upstream. 

This is also what the readme suggests:

> Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

None

### Hosting type

None

2.4.2

thomasheartman:
This PR updates how custom headers work. Previously, any custom headers you added would only get added to feature get/post requests. Now we also add them to metrics requests. This is in line with what the documentation implies and what we think it's reasonable to expect. As such, I consider this a bug fix.

## About the changes

The most relevant changes are in the `src/metrics.ts`. In short, we perform the same kind of method here as in `src/index.ts` to enrich the standard headers with any custom headers that are provided.

We also take care to pass the custom headers from the client constructor to the metrics constructor.

### Tests

The change comes with the following tests:

1. Verify that the custom headers are passed on from the client constructor to the metrics constructor
2. Verify that custom headers ...
    - are sent
    - override preset headers, but only if they have a valid value
    - that are set to `null` or `undefined` don't get sent.

## Discussion points

At the time of writing, the current set of changes are essentially just a copy-paste of what we were doing in the `index.ts` file. While it's tempting to refactor this and extract it, that also seems like a bit of overkill to me. As far as I'm aware, there's only two copies, and they behave slightly differently (different headers), so I think the duplication is accidental more than intentional. We could parameterize it into a shared function (and I'm happy to do that), but I'm not sure it's worth it.

Second, we currently allow setting headers that are empty or all-whitespace strings. This is consistent with how the `index.ts` file does it. We can't change the index file, but we don't have to copy this behavior. Do we want to allow empty strings? I'm leaning towards yes because it's less surprising, but I'm open to hearing arguments.

Finally, I also considered applying the custom headers to the `fetch` parameter that we send in (by wrapping it in another function). This would allow us to not change anything inside the `metrics.ts` file at all, but I decided against it because I thought it might be harder to reason about and to test. Again, though, happy to do it if we think it's better.

---

Closes #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ The Unleash SDK takes the following options:
 | fetch             | no | `window.fetch` or global `fetch` | Allows you to override the fetch implementation to use. Useful in Node.js environments where you can inject `node-fetch`                    | 
 | bootstrap         | no | `[]` | Allows you to bootstrap the cached feature toggle configuration.                                                                               | 
 | bootstrapOverride | no| `true` | Should the bootstrap automatically override cached data in the local-storage. Will only be used if bootstrap is not an empty array.     | 
-| headerName        | no| `Authorization` | Provides possiblity to specify custom header that is passed to Unleash / Unleash Proxy with the `clientKey` | 
-| customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used. | 
-| impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. | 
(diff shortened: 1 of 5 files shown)


Tymek:
Yes! Let's do that

PR #150, opened 2023-05-04
feat: report metrics for variants

kwasniew:
## About the changes

Main capability:
* adding reporting of metrics for variants

Changes:
* introducing SENT event for easier testing and integration with other tools. Same event is present in node.js SDK
* added metrics unit test for counting variants
* added component tests for the whole SDK client for metrics reporting
* formatting code

Next steps:
* Proxy/Edge should accumulate client SDK metrics

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,6 +64,7 @@ export const EVENTS = {
     READY: 'ready',
     UPDATE: 'update',
     IMPRESSION: 'impression',
+    SENT: 'sent',
 };
 
 const IMPRESSION_EVENTS = {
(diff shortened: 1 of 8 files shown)


src/metrics.ts
```diff
@@ -142,6 +156,19 @@ export default class Metrics {
         return true;
     }
 
+    public countVariant(name: string, variant: string): boolean {
```
kwasniew, line 159: similar to Node.js implementation

src/index.ts
```diff
@@ -217,6 +219,9 @@ export class UnleashClient extends TinyEmitter {
         const enabled = toggle?.enabled || false;
         const variant = toggle ? toggle.variant : defaultVariant;
 
+        if (variant.name) {
+            this.metrics.countVariant(toggleName, variant.name);
```
kwasniew, line 223: variant is counted in addition to regular count for enabled/disabled

PR #154, opened 2023-05-09
2.5.0

kwasniew:
## About the changes

Official (non-beta) release with variants metrics. Previously we had beta. This new version will be used to bump up other JS based SDK clients.


thomasheartman (approved):
Looks good to me! ✅

PR #155, opened 2023-06-12
Upgrade

Tymek:
## About the changes

- [x] Update CI to build also with latest Node versions
- [x] Update to Yarn
- [x] Unify NPM release scripts (use the manual workflow from unleash/unleash)
- [x] Bump dependencies

This will require **major version bump**, because of dropped Node v12 support.

Closes [#1-951](https://linear.app/unleash/issue/1-951/sdk-js-browser-client-maintenance)

### Important files
- rollup.config.mjs
- .github/workflows/release.yml


package.json
```diff
@@ -1,22 +1,19 @@
...
   ],
   "scripts": {
-    "preversion": "rm -rf build && npm run build",
-    "prepare": "npm run build && ./scripts/check-release.sh",
-    "build": "npm run build:ts && npm run build:web",
-    "build:ts": "tsc",
-    "build:web": "rollup -c rollup.config.js",
-    "fmt": "prettier src --write --loglevel warn",
-    "lint": "eslint src",
+    "build": "rm -rf build && tsc && rollup -c rollup.config.mjs",
```
gastonfournier, line 13: Nit: Here the build is doing a clean & build, which seems like two different steps, but for CI/CD it doesn't matter much because build folder will not exist

sjaanus (approved):
LG

tsconfig.json
```diff
@@ -4,15 +4,15 @@
...
+    "lib": ["es6"],                           /* Specify library files to be included in the compilation. */
     // "allowJs": true,                       /* Allow javascript files to be compiled. */
     // "checkJs": true,                       /* Report errors in .js files. */
     // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
     "declaration": true,                      /* Generates corresponding '.d.ts' file. */
     "declarationMap": true,                   /* Generates a sourcemap for each corresponding '.d.ts' file. */
     "sourceMap": true,                        /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    "outDir": "./build",                      /* Redirect output structure to the directory. */
+    "outDir": "build/cjs",                      /* Redirect output structure to the directory. */
```
gastonfournier, line 15: Have you tested importing this as a library? These changes tend to be tricky

gastonfournier (approved):
LG, but I'd recommend you to pack this as a library, install and test it from the consumer side

tsconfig.json
Tymek: Yes. Tested both in CJS and ESM setup (require & import)

package.json
Tymek: `rm -f` will not cause an error if the target doesn't exist

PR #170, opened 2023-10-09
chore: update dependencies

Tymek:
## About the changes
- Bump dependencies. Only significant major version change is "prettier", that changed default "trailingComma" setting.
- use ESModule


andreas-unleash (approved):
LGTM

Issue #173, opened 2023-10-14
CommonJS imports are broken in v3.0.0


gamedevsam:
### Describe the bug

My server suddenly stopped working during startup as a result of attempting to import `unleash-proxy-client` v3.0.0. I believe this is because `unleash-proxy-client` package defines `"type": "module"` in its `package.json` which is not something you should do in NPM packages (this field is for applications to define, not libraries, [more info here](https://nodejs.org/docs/v18.18.2/api/packages.html#type)).

### Steps to reproduce the bug

Create a new Node project, attempt to import unleash-proxy-client using `require`, and start the application. Notice the error, ex:

```ts
const unleash_proxy_client = require("unleash-proxy-client");
```

### Expected behavior

Using CommonJS to import the package should not result in a runtime error.

### Logs, error output, etc.

```bash
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/samuel/WSL_Dev/Personal/dokku_app/node_modules/.pnpm/unleash-proxy-client@3.0.0/node_modules/unleash-proxy-client/build/cjs/index.js from /home/samuel/WSL_Dev/Personal/dokku_app/src/server/dist/shared/modules/stores/feature_flags_store.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /home/samuel/WSL_Dev/Personal/dokku_app/node_modules/.pnpm/unleash-proxy-client@3.0.0/node_modules/unleash-proxy-client/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at Object.<anonymous> (/home/samuel/WSL_Dev/Personal/dokku_app/src/server/dist/shared/modules/stores/feature_flags_store.js:4:32)
    at Object.<anonymous> (/home/samuel/WSL_Dev/Personal/dokku_app/src/server/dist/server/src/boot_server.js:15:31)
    at /home/samuel/WSL_Dev/Personal/dokku_app/src/server/dist/server/src/main.js:29:43 {
  code: 'ERR_REQUIRE_ESM'
}
```

### Unleash version

3.0.0

### Subscription type

None

### Hosting type

None

sjaanus:
[...]

PR #174, opened 2023-10-16
fix: revert to commonjs

sjaanus:
Fixes https://github.com/Unleash/unleash-proxy-client-js/issues/173

We are not ready to support ESM yet. Reverting back to CommonJS, so we don't break build.


sighphyre (commented):
The change looks good, the tests seem to be upset tho

sighphyre (approved):
G

PR #175, opened 2023-10-16
fix: support also require

sjaanus:


Tymek (approved):
All good once you check Node v16 compatibility and revert unneeded yarn.lock changes 👍🏻

PR #183, opened 2023-11-14
feat: abort controller

kwasniew:
## About the changes

Abort controller that aborts previous request when a new one is made.

Sample use cases:
* slow network + multiple consecutive updateContext calls
* slow network + very short interval


src/index.test.ts
```diff
@@ -576,6 +576,25 @@ test('Should publish error when fetch fails', (done) => {
...
+        clientKey: '12',
+        appName: 'web',
+    };
+    const client = new UnleashClient(config);
+    await client.start();
+    client.updateContext({ userId: '123' }); // abort 1
+    client.updateContext({ userId: '456' }); // abort 2
+    await client.updateContext({ userId: '789' });
+
+    expect(abortSpy).toBeCalledTimes(2);
```
kwasniew, line 594: I also wanted to test error emitter on abort but the mocking fetch library does not support AbortController

src/index.ts
```diff
@@ -89,6 +90,18 @@ export const resolveFetch = () => {
     return undefined;
 };
 
+const resolveAbortController = () => {
```
kwasniew, line 93: similar pattern to how we handle fetch

sighphyre (approved):
Yes! This is very nice

src/index.ts
```diff
@@ -94,6 +95,18 @@ export const resolveFetch = () => {
...
+    try {
+        if (typeof window !== 'undefined' && 'AbortController' in window) {
+            return () => new window.AbortController();
+        } else if ('fetch' in globalThis) {
+            return () => new globalThis.AbortController();
+        }
+    } catch (e) {
+        console.error('Unleash failed to resolve "AbortController" factory', e);
```
sighphyre, line 106: This should basically never happen, right? This is if someone is doing something unusual like swapping out fetch?

kwasniew: Correct - someone would need to take deliberate effort to end up in this situation

nammi-kobler:
I keep getting these:

image

the requests comes in pairs (within 2 seconds), and if I upgrade to v3.3.0 I get these every other request (so one is always `cached` and the other has this error in the console:

image

downgrading to 3.2.0 fixes these for me.

this is what Stack Overflow tells me about the `NS_BINDING_ABORTED` ones:
https://stackoverflow.com/questions/704561/ns-binding-aborted-shown-in-firefox-with-httpfox

kwasniew:
@nammi-kobler those errors are expected when the requests are aborted e,g, when you're on a slow network and you have very short refreshInterval and the original request hasn't finished and you start the next one. From what I understand your refresh interval is 2s. Can you increase it and see if the problem persists? The abort operation should be an exception not a regular operation.

nammi-kobler:
The refresh interval is set to15s, but for some reason they always come in pairs (every 15 sec), and then one of them always fails. any idea on why they would come in pairs?

kwasniew:
@nammi-kobler Can you check if this PR https://github.com/Unleash/unleash-proxy-client-js/pull/207 fixes your issue? It's applied in the https://github.com/Unleash/unleash-proxy-client-js/releases/tag/v3.4.0 release

nammi-kobler:
> @nammi-kobler Can you check if this PR #207 fixes your issue? It's applied in the https://github.com/Unleash/unleash-proxy-client-js/releases/tag/v3.4.0 release

Thanks for this!
Firefox:
image

Vivaldi:
image

Safari:
image

Looks better, but it seems like it still happens over time

kwasniew:
@nammi-kobler Thanks for checking. So is it correct that the console.error is gone and now you're only left with the double request problem? Can you remind me if you use unleash-proxy-client-js directly or through react or nextjs wrapper? I couldn't reproduce this problem in this vanilla client so my guess is it's some wrapper adding those extra calls.

nammi-kobler:
> @nammi-kobler Thanks for checking. So is it correct that the console.error is gone and now you're only left with the double request problem? Can you remind me if you use unleash-proxy-client-js directly or through react or nextjs wrapper? I couldn't reproduce this problem in this vanilla client so my guess is it's some wrapper adding those extra calls.

we are using `unleash-proxy-client`

image

NextJS warning is from a wrapper that comes with Mantine I think?

kwasniew:
Are you running in Strict Mode by any chance? https://react.dev/reference/react/StrictMode
It runs code twice in development.

nammi-kobler:
> Are you running in Strict Mode by any chance? https://react.dev/reference/react/StrictMode It runs code twice in development.

haha yes. that's maybe the answer then.

PR #185, opened 2023-11-28
feat: fire event when error is present and http call succeeds

FredrikOseberg:
This PR adds a new event that fires when an error was recorded and the subsequent HTTP call is successful.

### Why
Our React Proxy SDK will set an error based on the error event fired from this SDK when the HTTP call fails. Without this event we have no way to effectively reset this error, because the only event that is fired consistently is the UPDATE event, which is only fired when the response is not 304. This does not cover the case when you intermittently lose connection and the next call succeeds with 304.

### How
We keep track of the sdkState internally in the SDK. When the HTTP calls error we change the state of the SDK to be in an error state. If the internal SDK state is set to error, and the HTTP call is successful we will emit a RECOVERED event that will allow subscribers to clear out stale errors and set the sdkState back to 'healthy'.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -67,7 +67,7 @@ export const EVENTS = {
     UPDATE: 'update',
     IMPRESSION: 'impression',
     SENT: 'sent',
-    POST_ERROR_SUCCESS: 'postErrorSuccess'
+    POST_ERROR_SUCCESS: 'postErrorSuccess',
 };
 
 const IMPRESSION_EVENTS = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -67,6 +67,7 @@ export const EVENTS = {
     UPDATE: 'update',
     IMPRESSION: 'impression',
     SENT: 'sent',
+    POST_ERROR_SUCCESS: 'postErrorSuccess'
 };
 
 const IMPRESSION_EVENTS = {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -177,11 +179,13 @@ export class UnleashClient extends TinyEmitter {
         this.refreshInterval = disableRefresh ? 0 : refreshInterval * 1000;
         this.context = { appName, environment, ...context };
         this.usePOSTrequests = usePOSTrequests;
+        this.sdkError = null;
```
kwasniew, line 182: what is we had a union with a current SDK state: type SdkState = 'healthy' | 'error' or similar?

gastonfournier: I like it and for the event name, may I suggest `RECOVERED` or `RECOVERED_FROM_ERROR`?

src/index.ts
```diff
@@ -81,6 +82,8 @@ const defaultVariant: IVariant = {
 };
 const storeKey = 'repo';
 
+type SdkState = 'initializing' | 'healthy' | 'error';
```
gastonfournier, line 85: Just another small detail: If I see my SDK state is error, should I trust its results? Maybe `stale` is more accurate to how our SDK behaves: if some request fails, we still respond correctly with potentially stale data.

gastonfournier (approved):

src/index.ts
FredrikOseberg: It's a private property though, so we only use it internally in the SDK to discern whether or not we should emit the recovered event. In this case I think it's actually more accurate for the reader to use 'error' and it's not exposed anywhere outside of the internals of the SDK.

PR #190, opened 2024-01-02
fix: don't send 'If-None-Match' header for empty etag

### LINKED ISSUE #189: Empty 'If-None-Match' header is send causing some firewalls to block the request
### Describe the bug

Requests are made with headers always containing the `If-None-Match` header, even if there is no value.

if no `etag` is set, it defaults to `''` and thus the `If-None-Match` header is set with an empty value.

Some firewalls, like "F5" are blocking requests with empty headers by default.

Don't set an etag

### Expected behavior

_No response_

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

None

### Hosting type

None

_No response_

sebkasanzew:
## About the changes

Closes #189 

### Important files

`src/index.ts`

## Discussion points

Should this only be applied to this `If-None-Match` header, or should maybe the `notNullOrUndefined` function in `./utils.ts` be also changed to `notNullOrUndefinedOrEmptyString`?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -380,8 +380,13 @@ export class UnleashClient extends TinyEmitter {
             [this.headerName]: this.clientKey,
             Accept: 'application/json',
             'Content-Type': 'application/json',
-            'If-None-Match': this.etag,
         };
+
+        // only add etag, if it's not an empty string
+        if (this.etag) {
+            headers['If-None-Match'] = this.etag;
+        }
+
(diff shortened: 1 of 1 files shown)


Tymek:
Done in #187 and released in [v3.3](https://github.com/Unleash/unleash-proxy-client-js/releases/tag/v3.3.0)

Issue #194, opened 2024-01-12
Missing `currentTime` in Unleash Context object


AdrienFromToulouse:
### Describe the bug

The `currentTime` property https://docs.getunleash.io/reference/unleash-context seems to be missing.

https://github.com/Unleash/unleash-proxy-client-js/blob/main/src/index.ts#L9

Cheers!

### Steps to reproduce the bug

Try to set the `currentTime` property on the context, it would result in noop.

### Expected behavior

Take the `currentTime` property into account

### Unleash version

3.2.0

### Subscription type

Pro

### Hosting type

Hosted by Unleash

### SDK information (language and version)

"@unleash/proxy-client-react": "^4.1.1"

sjaanus:
This is released as part of https://github.com/Unleash/unleash-proxy-client-js/releases/tag/v3.3.1

AdrienFromToulouse:
Thanks!

AdrienFromToulouse:
@sjaanus will it be incorporated to https://github.com/Unleash/proxy-client-react anytime soon?

Cheers,

PR #195, opened 2024-01-15
fix: add curent time to context defined fields

sjaanus:
Solves https://github.com/Unleash/unleash-proxy-client-js/issues/194


sighphyre (commented):
What does this do in practise? Can we get a test asserting the behaviour does what it says (dunno what that behaviour is from looking at this change)

sjaanus:
> What does this do in practise? Can we get a test asserting the behaviour does what it says (dunno what that behaviour is from looking at this change)

Yep, somehow test did not get into commit.

sighphyre (approved):
Thank you! That looks nicer!

PR #200, opened 2024-04-24
Fix: ready not emitted when bootstrap is set and bootrstrapOverride is false

### LINKED ISSUE #199: The READY event is not triggered when the bootstrap is not empty and bootstrapOverride = false.
### Describe the bug

The issue is as follows: The READY event is not triggered when the bootstrap is not empty and bootstrapOverride is set to false. 

For example, in the following code snippet, the Ready event will never be triggered. I can use the Init event, but I specifically need the Ready trigger event for other reasons. 

My request is for Events.Ready to emit in the example code snippet below:

```
this.unleash = new UnleashClient({
    url: '...',
    clientKey: '...',
    appName: 'frontend',
    refreshInterval: 3,
    environment: '...',
    context: {
        properties: {
            ....
        },
    },
    bootstrap: [
        {
            enabled: false,
            name: 'test-frontend',
            ...
        },
    ], 
    bootstrapOverride: false,
});
```

The Ready event is only triggered in two cases.

image

image

As you can see, these two code conditions do not cover the case where the bootstrap is not empty and bootstrapOverride is false.  (My toggles.length is not equal to 0)

### Expected behavior

The READY event needs to be triggered also when the bootstrap is not empty and bootstrapOverride = false.

andreas-unleash:
Fixes a bug where the Ready event was not emitted when bootstrap is defined and bootstrap override is false

Closes #199

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1675,20 +1675,20 @@ describe('READY event emission', () => {
         clientKey: '12',
         appName: 'web',
         bootstrap: [
-        {
-            enabled: false,
-            name: 'test-frontend',
-            variant: { name: 'some-variant', enabled: false},
-            impressionData: false
-        },
-    ],
+            {
+                enabled: false,
+                name: 'test-frontend',
+                variant: { name: 'some-variant', enabled: false },
+                impressionData: false,
+            },
+        ],
         bootstrapOverride: false,
         fetch: async () => {
             return {
                 ok: true,
                 headers: new Map(),
                 async json() {
-                    return {  };
+                    return {};
                 },
             };
         },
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -1666,3 +1666,52 @@ test('Should set sdkState to healthy when client is started', (done) => {
...
+        appName: 'web',
+        bootstrap: [
+            {
+                enabled: false,
+                name: 'test-frontend',
+                variant: { name: 'some-variant', enabled: false },
+                impressionData: false,
+            },
+        ],
+        bootstrapOverride: false,
```
gastonfournier, line 1685: Is this a new configuration? Not sure if boostrap override is a name that brings a lot of clarity... but maybe it's an existing setting

src/index.ts
```diff
@@ -466,7 +466,11 @@ export class UnleashClient extends TinyEmitter {
                         this.sdkState = 'healthy';
                     }
 
-                    if (!this.bootstrap && !this.readyEventEmitted) {
+                    if (
+                        (!this.bootstrap ||
+                            (this.bootstrap && !this.bootstrapOverride)) &&
+                        !this.readyEventEmitted
+                    ) {
```
gastonfournier, line 473: What would be the problem with just:
```suggestion
                    if (!this.readyEventEmitted) {
```

src/index.test.ts
andreas-unleash: No, it is an option that was there

src/index.ts
andreas-unleash: This covers the missing condition, your suggestion makes sense afaict but it is broader.

ivarconr: I simplified.

ivarconr (approved):

ivarconr (approved):

PR #202, opened 2024-05-06
feat: prevent fetching toggles when they are up to date

### LINKED ISSUE #201: Service worker compatibility
### Describe the feature request

We would like to use `unleash-proxy-client-js` in a service worker context, more precisely in the context of a browser extension service worker (and "non persistent background").

### Background

We are facing some issues while running `unleash-proxy-client-js` in a service worker context (browser extension service worker).

First issue : each time the proxy is instantiated, it fetches the flags, no matter how old they are in the storage. In the context of a service worker which stops and restarts multiple times depending on its use this will cause a huge overhead on our proxy server (we have hundreds of thousands of live extensions).

Second issue : the internal use of `setTimeout` and `setInterval` which is not compatible with the lifecycle of this kind of service workers which have to use "browser alarms" instead, see [documentation](https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers#convert-timers).

### Solution suggestions

For the first issue (fetch call on load) the solution could be to check "flag expiration" on load, if flags are younger than the "refreshInterval", it should not fetch (we have to store a new "last update" timestamp) OR use the `bootstrap` value on first load and don't fetch (wait for next refresh).

For the second issue (usage of `setInterval`), we could expose as public methods the "refresh" toggles and metrics so that library consumers can call them themself if needed (we could so add our own "alarm" to refresh the toggles and metrics).

jeremiewtd:
## About the changes

This pull request is a proposal for fixing this [issue](https://github.com/Unleash/unleash-proxy-client-js/issues/201) to add compatibility with service worker and browser extensions to avoid multiple requests to the proxy server each time the service worker is started.

It introduces a new option `togglesStorageTTL` that prevents fetching the flags if they are up to date (not older than the number of seconds defined in this option).

A second pull request will come to fix the "refresh" problem described in the issue.

Closes #201

### Important files

Main changes concerned `index.ts` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The Unleash SDK takes the following options:
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used if it is not `null` or `undefined`. `customHeaders` values that are `null` or `undefined` will be ignored. |
 | impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. |
 | environment | no | `default` | Sets the `environment` option of the [Unleash context](https://docs.getunleash.io/reference/unleash-context). This does **not** affect the SDK's [Unleash environment](https://docs.getunleash.io/reference/environments). |
-| togglesStorageTTL   | no | `0` | How much time (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched again. If set to 0 will disable checking for expiration and considered always expired                 |
+| togglesStorageTTL   | no | `0` | How long (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched again. If set to 0 will disable expiration checking and will be considered always expired.                 |
 
 ### Listen for updates via the EventEmitter
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ if (variant.name === 'blue') {
 
 #### Updating the Unleash context
 
-The [Unleash context](https://docs.getunleash.io/reference/unleash-context) is used to evaluate features against attributes of a the current user. To update and configure the Unleash context in this SDK, use the `updateContext` and `setContextField` methods.
+The [Unleash context](https://docs.getunleash.io/reference/unleash-context) is used to evaluate features against attributes of a the current user. To update and configure the Unleash context in this SDK, use the `updateContext`, `setContextField` and `removeContextField` methods.
 
 The context you set in your app will be passed along to the Unleash proxy or the front-end API as query parameters for feature evaluation.
 
@@ -101,6 +101,14 @@ The `setContextField` method only acts on the property that you choose. It does 
 unleash.updateContext({ userId: '1233' });
 
 unleash.setContextField('userId', '4141');
+```
+
+The `removeContextField` method only acts on the property that you choose. It does not affect any other properties of the Unleash context.
+
+```js
+unleash.updateContext({ userId: '1233', sessionId: 'sessionNotAffected' });
+
+unleash.removeContextField('userId');
 ```
 
 ### Available options
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ The Unleash SDK takes the following options:
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used if it is not `null` or `undefined`. `customHeaders` values that are `null` or `undefined` will be ignored. |
 | impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. |
 | environment | no | `default` | Sets the `environment` option of the [Unleash context](https://docs.getunleash.io/reference/unleash-context). This does **not** affect the SDK's [Unleash environment](https://docs.getunleash.io/reference/environments). |
+| togglesStorageTTL   | no | `0` | How much time, in seconds, the toggles are considered valid and should not be fetched again. If set to 0 will disable checking for expiration and considered always expired                 |
 
 ### Listen for updates via the EventEmitter
 
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The Unleash SDK takes the following options:
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used if it is not `null` or `undefined`. `customHeaders` values that are `null` or `undefined` will be ignored. |
 | impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. |
 | environment | no | `default` | Sets the `environment` option of the [Unleash context](https://docs.getunleash.io/reference/unleash-context). This does **not** affect the SDK's [Unleash environment](https://docs.getunleash.io/reference/environments). |
-| togglesStorageTTL   | no | `0` | How much time, in seconds, the toggles are considered valid and should not be fetched again. If set to 0 will disable checking for expiration and considered always expired                 |
+| togglesStorageTTL   | no | `0` | How much time (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched again. If set to 0 will disable checking for expiration and considered always expired                 |
 
 ### Listen for updates via the EventEmitter
 
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ The Unleash SDK takes the following options:
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used if it is not `null` or `undefined`. `customHeaders` values that are `null` or `undefined` will be ignored. |
 | impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. |
 | environment | no | `default` | Sets the `environment` option of the [Unleash context](https://docs.getunleash.io/reference/unleash-context). This does **not** affect the SDK's [Unleash environment](https://docs.getunleash.io/reference/environments). |
-| togglesStorageTTL   | no | `0` | How long (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched again. If set to 0 will disable expiration checking and will be considered always expired.                 |
+| togglesStorageTTL   | no | `0` | How long (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched on start. If set to 0 will disable expiration checking and will be considered always expired.                 |
 
 ### Listen for updates via the EventEmitter
 
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -703,9 +703,7 @@ test('Should include etag in second request', async () => {
     const firstRequest = getTypeSafeRequest(fetchMock, 0);
     const secondRequest = getTypeSafeRequest(fetchMock, 1);
 
-    expect(firstRequest.headers).toMatchObject({
-        'If-None-Match': '',
-    });
+    expect(firstRequest.headers).toMatchObject({});
     expect(secondRequest.headers).toMatchObject({
         'If-None-Match': etag,
     });
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -60,6 +60,9 @@ test('Should perform an initial fetch as POST', async () => {
 
     expect(request.method).toBe('POST');
     expect(body.context.appName).toBe('webAsPOST');
+    expect(request.headers).toMatchObject({
+        'Content-Type': 'application/json',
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -132,6 +132,7 @@ The Unleash SDK takes the following options:
 | customHeaders     | no| `{}` | Additional headers to use when making HTTP requests to the Unleash proxy. In case of name collisions with the default headers, the `customHeaders` value will be used if it is not `null` or `undefined`. `customHeaders` values that are `null` or `undefined` will be ignored. |
 | impressionDataAll | no| `false` | Allows you to trigger "impression" events for **all** `getToggle` and `getVariant` invocations. This is particularly useful for "disabled" feature toggles that are not visible to frontend SDKs. |
 | environment | no | `default` | Sets the `environment` option of the [Unleash context](https://docs.getunleash.io/reference/unleash-context). This does **not** affect the SDK's [Unleash environment](https://docs.getunleash.io/reference/environments). |
+| togglesStorageTTL   | no | `0` | How long (Time To Live), in seconds, the toggles in storage are considered valid and should not be fetched again. If set to 0 will disable expiration checking and will be considered always expired.                 |
```
ivarconr, line 135: Is there any reason we would not default this to the `refreshInterval`? 

I can see the argument of having it "opt-in" initially, until we validate the behavior, but it would be great win for everyone to not overload the Unleash Frontend API (or Edge) if a user decides to force-reload multiple times?

src/index.ts
```diff
@@ -423,8 +432,29 @@ export class UnleashClient extends TinyEmitter {
         await this.storage.save(storeKey, toggles);
     }
 
+    private isUpToDate() {
+        if (this.togglesStorageTTL <= 0 || this.toggles.length === 0) {
```
ivarconr, line 436: This will still hammer the API if the toggles length is actually is 0, which can happen in some cases:

- A fresh Unleash installation
- you are connecting it to a empty project

src/index.ts
```diff
@@ -423,8 +432,29 @@ export class UnleashClient extends TinyEmitter {
...
+    private async updateLastRefresh() {
+        this.lastRefreshTimestamp = Date.now();
+        this.storage.save(lastUpdateKey, this.lastRefreshTimestamp);
+    }
+
     private async fetchToggles() {
         if (this.fetch) {
+            if (this.isUpToDate()) {
+                this.emitReady();
+                return;
```
ivarconr, line 456: This method have been carefully designed to not use "early returns". Early return in combination with try/catch control flow increases the complexity. I am not comfortable about breaking the current pattern to support avoiding the initial fetch use case.

I think we instead can have a specific `initalFetchMethod`, that only calls `fetchToggles` if the cached data is not considered recent.

src/index.ts
```diff
@@ -479,7 +496,21 @@ export class UnleashClient extends TinyEmitter {
...
+                    await this.storeToggles(data.toggles);
+
+                    if (this.sdkState !== 'healthy') {
+                        this.sdkState = 'healthy';
+                    }
+
+                    this.emitReady();
+                }
+
+                this.updateLastRefresh();
```
ivarconr, line 513: This should happen when we get a positive response from the server (2xx or 304).

src/index.ts
```diff
@@ -489,6 +520,13 @@ export class UnleashClient extends TinyEmitter {
             }
         }
     }
+
+    private emitReady() {
```
ivarconr, line 524: its a great improvement to have this in its on little function as it is duplicated a few places! 👍🏼

src/index.test.ts
```diff
@@ -1483,15 +1555,16 @@ test('Should emit impression events on getVariant calls when impressionData is f
 });
 
 test('Should publish ready only when the first fetch was successful', async () => {
+    expect.assertions(3);
```
ivarconr, line 1558: its a bit confusing to see updates to existing tests that seems unrelated to the PR. Can you please elaborate on why you needed to modify this scenario?

ivarconr: (the change looks reasonable tbh, but still would love to see your comments on them)

src/index.ts
```diff
@@ -92,6 +93,7 @@ const defaultVariant: IVariant = {
     feature_enabled: false,
 };
 const storeKey = 'repo';
+const lastUpdateKey = 'lastUpdate';
```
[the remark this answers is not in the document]
ivarconr: I think the naming could be slightly better. Would love more opinions on the name.

src/index.ts
ivarconr: By running this on each fetch we have another potential bug in the future: What if the stored timestamp somehow ends up far in to the future (we are dealing with untrusted browsers) we will never try to fetch updates at all.

README.md
jeremiewtd: > Is there any reason we would not default this to the refreshInterval?

This is because in our case, we don't want to use the `refreshInterval` because it uses `setInterval` that we can't use in our "browser extension" context. Moreover if anyone doesn't want the refresh feature to be enabled, they will not be able to use this new feature if there is no option for it.

> I can see the argument of having it "opt-in" initially, until we validate the behavior, but it would be great win for everyone to not overload the Unleash Frontend API (or Edge) if a user decides to force-reload multiple times?

Do you mean this should be enabled by default?

src/index.ts
jeremiewtd: Ok we can remove that, we thought we didn't want to set empty toggles in the cache but I see your point.

src/index.ts
jeremiewtd: Yes you're right, the `initialFetchMethod` is a better idea than this, we will rework it.

Come to think of it, this code is actually buggy... because if we update the context and the `isUpToDate` function says "yes", the flags will not be updated, which is not intended.

src/index.ts
jeremiewtd: yes this is the case no?

src/index.test.ts
jeremiewtd: In an early version of our implementation, we needed to update this test and actually we realized that the test was not working as it should so we fixed it, but we can remove that if you think that should not be part of this pull request.

src/index.ts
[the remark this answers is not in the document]
jeremiewtd: We were not very inspired for this name 😃, I think a combination of the two proposals could be the best?
```suggestion
const lastUpdateKey = 'repoLastUpdateTimestamp';
```

jeremiewtd:
> * we need to define the behaviour when bootstrap is used together with `bootstrapOverride` property. Should we then assume the bootstrap is fresh and avoid the initial fetch? Should we keep the stored timestamp but still override with the store with the bootstrapped config? Do we need an additional option for this? For context: Bootstrap is often used to hydrate the SDK when the app has been part of a serverside render and flags was resolved as part of the serverside render (to avoid flickering).

I think as you said we could "assume the bootstrap is fresh and avoid the initial fetch" and so update the timestamp too.
Are you ok with this?

jeremiewtd (commented):
We made some changes based on your review suggestions, we are waiting for your answers to the remaining questions before doing anything else.

jeremiewtd:
Come to think of it, should the context be part of the timestamp "key"? Otherwise the saved toggles could be wrong if they were saved in a different context than the current starting one in a new instance?

src/index.test.ts
ivarconr: thanks for a clarification. I would prefer to have it as a separate PR if you do not mind. We can keep it here as well

ivarconr:
> Come to think of it, should the context be part of the timestamp "key"? Otherwise the saved toggles could be wrong if they were saved in a different context than the current starting one in a new instance?

This makes me think we could store the etag as well. This would yield a validation call to the API but return a 304 if nothing changed. In terms of Unleash/Edge that is a lot cheaper than a full 200 response. **Would that be good enough, or do you still see the need to avoid the initial HTTP call?** 

If we keep the TTL approach we should definitely store a hash of the context object, to make sure the context has not changed in between.

jeremiewtd:
> > Come to think of it, should the context be part of the timestamp "key"? Otherwise the saved toggles could be wrong if they were saved in a different context than the current starting one in a new instance?
> 
> This makes me think we could store the etag as well. This would yield a validation call to the API but return a 304 if nothing changed. In terms of Unleash/Edge that is a lot cheaper than a full 200 response. **Would that be good enough, or do you still see the need to avoid the initial HTTP call?**
> 
> If we keep the TTL approach we should definitely store a hash of the context object, to make sure the context has not changed in between.

The Etag seems like a good idea in absolute terms but it doesn't solve our main problem which is the number of calls to the server. Do you think we should continue with this PR or do you prefer to only use the etag?

ivarconr:
> The Etag seems like a good idea in absolute terms but it doesn't solve our main problem which is the number of calls to the server. Do you think we should continue with this PR or do you prefer to only use the etag?

When I think about it feels like two different needs. I would like to keep the TTL separate. In order to test this feature more widely before we commit I would also like to have the option as an experimental option and not part of the official API just yet.

src/index.ts
```diff
@@ -53,6 +53,7 @@ interface IConfig extends IStaticContext {
     customHeaders?: Record<string, string>;
     impressionDataAll?: boolean;
     usePOSTrequests?: boolean;
+    togglesStorageTTL?: number;
```
ivarconr, line 56: Instead of adding it directly to the config definition, If we could have an optional "experimental" container we could make it clear that this option might go away or be changed in the future. 

```
experimental?: {
  togglesStorageTTL?: number;
}
```

jeremiewtd: Ok we are working on it and adding the hashed context. Do you have any proposal for how you would like this to be done?

Florent-Wanteeed: Hello

Concerning the hashed context, we started with the use of this type of library : “[object-hash](https://www.npmjs.com/package/object-hash)”. 
Is this OK for you ?

“[unleash-proxy-client](https://bundlephobia.com/package/unleash-proxy-client@3.4.0)” lib size => 5.5 KB 
“[object-hash](https://bundlephobia.com/package/object-hash@3.0.0)” lib size => 10 KB

Which is a significant increase in percentage terms.

Last lib update: 02/2022 ([v3.0.0](https://bundlephobia.com/package/object-hash@3.0.0)), 

Otherwise we can simply use a JSON.stringify (more verbose but which does not depend on anything)

src/index.ts
ivarconr: Anything in user-land cannot be trusted. Users might suddenly change their local time etc, which forces us to be a bit restrictive on what we trust. E.g. if the stored data is 2 hours in the future we should probably disregard the cached data and do a fetch.

src/index.ts
ivarconr: We do not want to take on a library of this size for this. 

I think the JSON.stringify is a good start, and as this is an experimental feature we can use that to learn the performance impact here. In the future we can consider to take advantage of the built in hasing capabilities, supported by most modern browsers (https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest)

ivarconr:
Oh, there are a few lint issues. Do you mind running "yarn format" on it to automatically fix these @Florent-Wanteeed @jeremiewtd ?

src/util.ts
```diff
@@ -26,3 +26,24 @@ export const urlWithContextAsQuery = (url: URL, context: IContext) => {
...
+            sortedObj[key] = sortObjectProperties(
+                obj[key] as Record<string, unknown>
+            );
+        } else {
+            sortedObj[key] = obj[key];
+        }
+    });
+
+    return sortedObj;
+};
```
Tymek, line 46: Context has a simple structure. I don't think we need recursion here.
https://docs.getunleash.io/reference/unleash-context#structure

Output doesn't need to directly correspond to context structure.

```ts
const sortEntries = (record: Record<string, string>) => {
    return Object.entries(record).sort(([a], [b]) => a.localeCompare(b));
};

const contextString = (context: IContext): string => {
    const { properties, ...fields } = context;

    return JSON.stringify({
        fields: sortEntries(fields),
        properties: sortEntries(context.properties || {}),
    });
};
```

src/util.test.ts
```diff
@@ -57,3 +58,23 @@ test('should exclude context properties that are null or undefined', async () =>
...
+
+    test('Should compute hash value for an object with not sorted keys', () => {
+        const obj: IContext = { userId: '4', appName: '1', environment: '3', currentTime: '2' };
+        const hashValue = computeContextHashValue(obj);
+        expect(hashValue).toBe('{"appName":"1","currentTime":"2","environment":"3","userId":"4"}');
+    });
+
+    test('Should compute hash value for an object with nested objects and not sorted keys', () => {
+        const obj: IContext = { appName: '1', properties: { d: "4", c: "3" }, currentTime: '2' };
+        const hashValue = computeContextHashValue(obj);
```
Tymek, line 77: I'd like to what the output is when parsing date

src/index.ts
```diff
@@ -476,33 +576,37 @@ export class UnleashClient extends TinyEmitter {
...
+                            type: 'HttpError',
+                            code: response.status,
+                        });
+
+                        this.lastError = {
+                            type: 'HttpError',
+                            code: response.status,
+                        };
+                    }
+                } else {
```
Tymek, line 597: Success-first statement is easier to read here, because there is no early return.

```
if (response.ok) {
} else if (response.status === 304) {
} else {}
```

(Using the fact that [`response.ok` is a shorthand](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok))

Tymek (commented):
Neat! I'll test it in practical setup. What I also would like to see is use of proper hashing. Better not to persist PII when there is no need to do it. `JSON.stringify` can be a fallback.

Tymek:
Merged in https://github.com/Unleash/unleash-proxy-client-js/pull/224

PR #209, opened 2024-05-21
test: aborted requests should not trigger error state

kwasniew:
## About the changes

Tests for https://github.com/Unleash/unleash-proxy-client-js/pull/207. Abort controller should not trigger errors.


FredrikOseberg (approved):
LGTM

PR #214, opened 2024-06-14
feat: allow access to sdk state

Tymek:
## About the changes
Push model for state changes is not enough when passing client to React SDK,
see https://github.com/Unleash/proxy-client-react/issues/168

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -150,6 +150,7 @@ export class UnleashClient extends TinyEmitter {
     private usePOSTrequests = false;
     private started = false;
     private sdkState: SdkState;
+    private lastError: any;
 
     constructor({
         storageProvider,
@@ -206,6 +207,7 @@ export class UnleashClient extends TinyEmitter {
                     console.error(error);
                     this.sdkState = 'error';
                     this.emit(EVENTS.ERROR, error);
+                    this.lastError = error;
                     resolve();
                 });
         });
@@ -393,6 +395,14 @@ export class UnleashClient extends TinyEmitter {
         this.metrics.stop();
     }
 
+    public isReady(): boolean {
+        return this.readyEventEmitted;
+    }
+
+    public getError(): SdkState {
+        return this.sdkState === 'error' ? this.lastError : undefined;
+    }
+
     private async resolveSessionId(): Promise<string> {
         if (this.context.sessionId) {
             return this.context.sessionId;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -291,3 +291,30 @@ const unleash = new UnleashClient({
...
+      stopped-->starting;
+```
+
+- `created` - The initial state of the client
+- `initialized` - The client has read configuration and local cached data
+- `bootstrapped` - SDK is configured to work offline and has been bootstrapped with feature toggles
+- `starting` - flags where not updated from Unleash yet
+- `healthy` - all flags are up to date
+- `stopped` - the client has been stopped manually and will not check for updates
+- `error` - an error has occurred while loading flags
```
gastonfournier, line 318: I find this a bit confusing and I believe we should align on these states as all our proxy-sdks should behave more or less the same way, so I believe this change is an important decision.

README.md
```diff
@@ -291,3 +291,21 @@ const unleash = new UnleashClient({
...
+
+```mermaid
+  graph LR;
+      created-->initialized;
+      created-->|READY|bootstrapped;
+      initialized-->starting;
+      bootstrapped-->healthy;
+      starting-->|READY|healthy;
+      healthy-->stopped;
+      stopped-->starting;
```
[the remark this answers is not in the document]
Tymek: 1. I added a bit more text. This is a fairly complex topic.
2. There is no external trigger for most of the transitions (except for stopping the client). Internal state depends on initial config of the client and ingested data
3. It depends. In some cases we throw Error (incompatible URL or no client key)

src/index.ts
```diff
@@ -393,6 +395,14 @@ export class UnleashClient extends TinyEmitter {
         this.metrics.stop();
     }
 
+    public isReady(): boolean {
+        return this.readyEventEmitted;
+    }
+
+    public getError(): SdkState {
```
ivarconr, line 402: can we rename to "getLastError"?

ivarconr (approved):
One method rename suggested. Other than that this looks good.

src/index.ts
Tymek: It has the logic to not return it when current SDK state is not an error. `getLastError` in my mind indicates that it returns last error even if SDK already recovered

ivarconr: ok, thanks for clarifying. makes sense.

PR #217, opened 2024-06-24
Fix: emit READY event once when bootstrapping

Tymek:
## About the changes
_READY_ event was emitted twice if bootstrapping (and overriding cache: `bootstrapOverride` `true`, that is the default)


Tymek:
@ivarconr I'm not sure if this SDK behavior is intentional or not

ivarconr:
Ready should only be emitted after a successful fetch from the Unleash API.

PR #219, opened 2024-06-26
Update README.md

Tymek:
Document behavior of READY event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ unleash.on('update', () => {
 
 - **error** - emitted when an error occurs on init, or when fetch function fails, or when fetch receives a non-ok response object. The error object is sent as payload.
 - **initialized** - emitted after the SDK has read local cached data in the storageProvider. 
-- **ready** - emitted after the SDK has successfully started and performed the initial fetch towards the Unleash Proxy. When bootstrapping client can emit this event twice - once when bootstrapped flags are loaded, and once on first connection to Unleash.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,7 +154,7 @@ unleash.on('update', () => {
 
 - **error** - emitted when an error occurs on init, or when fetch function fails, or when fetch receives a non-ok response object. The error object is sent as payload.
 - **initialized** - emitted after the SDK has read local cached data in the storageProvider. 
-- **ready** - emitted after the SDK has successfully started and performed the initial fetch towards the Unleash Proxy. 
+- **ready** - emitted after the SDK has successfully started and performed the initial fetch towards the Unleash Proxy. When bootstrapping client can emit this event twice - once when bootstrapped flags are loaded, and once on first connection to Unleash.
```
thomasheartman, line 157: Couple questions here:

You say the client *can* emit the event twice. Is that a "maybe" or is it that it *will*?

Oh, and we should probably updated it to mention Edge or the front-end API too.

```suggestion
- **ready** - emitted after the SDK has successfully started and performed the initial fetch of flags via the network (Edge, proxy, or front-end API). When bootstrapping, the client can emit this event twice: once when the bootstrapped flags are loaded, and once on first successful connection to Unleash.
```

Tymek: "can" as in "it will if both bootstrapping data is correct _and_ it is able to connect to upstream"

thomasheartman: got it. it's probably clear enough, then.

thomasheartman (approved):

PR #224, opened 2024-08-05
feat: prevent fetch on load

Tymek:
## About the changes
Copy of PR #202, with changes making it ready to be merged into main.

Diff: https://github.com/wanteeed/unleash-proxy-client-js/compare/feature/prevent-fetch-on-load...Unleash:unleash-proxy-client-js:feature/prevent-fetch-on-load

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -27,25 +27,44 @@ export const urlWithContextAsQuery = (url: URL, context: IContext) => {
     return urlWithQuery;
 };
 
-const sortObjectProperties = (
-    obj: Record<string, unknown>
-): Record<string, unknown> => {
-    const sortedKeys = Object.keys(obj).sort();
-    const sortedObj: Record<string, unknown> = {};
-    sortedKeys.forEach((key) => {
-        if (obj[key] !== null && typeof obj[key] === 'object') {
-            sortedObj[key] = sortObjectProperties(
-                obj[key] as Record<string, unknown>
-            );
-        } else {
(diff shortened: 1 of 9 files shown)


src/util.ts
```diff
@@ -26,3 +26,42 @@ export const urlWithContextAsQuery = (url: URL, context: IContext) => {
...
+
+    const sortEntries = (record: Record<string, string>) =>
+        Object.entries(record).sort(([a], [b]) =>
+            a.localeCompare(b, undefined)
+        );
+
+    return JSON.stringify([sortEntries(fields), sortEntries(properties)]);
+};
+
+const sha256 = async (input: string): Promise<string> => {
```
kwasniew, line 41: I'd use a non-cryptographic hash function suitable for general hash-based lookup e.g. murmurhash that we already use in the node client. It should be faster than cryptography hashing, there's a tiny library we already use elsewhere and it's tested.

kwasniew: Also I've just found that we're hashing only the context which is usually quite small so maybe we can start with no hashing in the first version?

Tymek: Yes, we can skip it in this version, but I'd like to have it to prevent storing PII in Unleash storage.

Speed difference is negligible  (~5.5ms for murmur vs ~6.2ms for SHA-256, for medium size context). I don't think it's worth additional weight (https://bundlephobia.com/package/murmurhash3js@3.0.1)

kwasniew (approved):

PR #229, opened 2024-09-04
fix: make sessionId be a string when set automatically

chriswk:
So, all of our schemas (even our type definitions) define sessionId to be a string, however, when user does not set one, we generate one automatically as a number, and never convert it back to a string. Don't really know how Typescript allowed this to pass. I guess since we only write the id to localstorage and don't really directly return the generated number, it threw its hands up in the air and gave up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,5 +86,6 @@
     "rules": {
       "@typescript-eslint/no-explicit-any": "off"
     }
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
Thank you!

PR #235, opened 2025-01-07
feat(1 3223): add client identification headers

thomasheartman:
This PR adds client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
- `x-unleash-appname`: the name of the application that is using the client
- `x-unleash-connection-id`: a unique identifier for the current instance of the client
- `x-unleash-sdk`: sdk information in the format `unleash-js@<version>`

## Discussion points:

What should we call the client here? Is `unleash-js` correct? Should it be `unleash-browser`? `unleash-proxy-client`?

Second: should the client be possible to override? This package is also what powers our React, Svelte, and Vue SDKs. Should they be able to override it (so you'd get `unleash-react@<version>`)? I think that would be useful

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,4 +1,5 @@
 import { FetchMock } from 'jest-fetch-mock';
+import { version as sdkVersion } from '../package.json';
 import 'jest-localstorage-mock';
 import * as data from './test/testdata.json';
 import IStorageProvider from './storage-provider';
@@ -1372,7 +1373,7 @@ test('Should add X-UNLEASH headers', async () => {
         /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
     expect(request.headers).toMatchObject({
-        'x-unleash-sdk': 'unleash-js@1.0.0',
+        'x-unleash-sdk': `unleash-js@${sdkVersion}`,
         'x-unleash-connection-id': expect.stringMatching(uuidFormat),
         'x-unleash-appname': 'web',
     });
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1356,6 +1356,28 @@ test('Should pass custom headers', async () => {
     });
 });
 
+test('Should add X-UNLEASH headers', async () => {
+    fetchMock.mockResponses([JSON.stringify(data), { status: 200 }]);
+    const config: IConfig = {
+        url: 'http://localhost/test',
+        clientKey: 'some123key',
+        appName: 'web',
+    };
+    const client = new UnleashClient(config);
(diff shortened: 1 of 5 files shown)


src/index.test.ts
```diff
@@ -1356,6 +1357,46 @@ test('Should pass custom headers', async () => {
...
+        /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+    const expectedHeaders = {
+        'x-unleash-sdk': `unleash-js@${sdkVersion}`,
+        'x-unleash-connection-id': expect.stringMatching(uuidFormat),
+        'x-unleash-appname': appName,
+    };
+
+    const getConnectionId = (request: any) =>
+        JSON.parse(JSON.stringify(request.headers))['x-unleash-connection-id'];
```
kwasniew, line 1391: would `const getConnectionId = (request: any) => request.headers['x-unleash-connection-id'];` work?

src/metrics.ts
```diff
@@ -121,6 +129,9 @@ export default class Metrics {
             [this.headerName]: this.clientKey,
             Accept: 'application/json',
             'Content-Type': 'application/json',
+            'x-unleash-sdk': `unleash-js@${this.sdkVersion}`,
```
kwasniew, line 132: we can start with just unleash-js even when used in other SDKs like React or Vue. Over time we can make it configurable if we have a good use case for it.

kwasniew (approved):

src/index.test.ts
thomasheartman: I hoped it would, but no such luck when I tried yesterday. I'll try again.

src/metrics.ts
thomasheartman: Yeah, cool. I ilke that.

src/index.test.ts
```diff
@@ -1356,6 +1357,46 @@ test('Should pass custom headers', async () => {
     });
 });
 
+test('Should add X-UNLEASH headers', async () => {
```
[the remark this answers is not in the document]
thomasheartman: Sure. Will fix.

src/index.test.ts
thomasheartman: Huh, it worked in the function, but not when applied to the header directly. My guess is that because we say it's `any`, then we lose any checking that would prevent you from getting the header.

It didn't work yesterday because trying to do `featureRequest.headers['x-unleash-connection-id']` gives you this error:

```
Element implicitly has an 'any' type because expression of type '"x-unleash-connection-id"' can't be used to index type 'HeadersInit'.
      Property 'x-unleash-connection-id' does not exist on type 'HeadersInit'.
```

PR #237, opened 2025-01-15
fix: embed version at compile time

kwasniew:
## About the changes

Removes requiring version at runtime with require: https://github.com/Unleash/unleash-proxy-client-js/commit/3615f15dd4e297f0f82ea17eeec6f9791cf16f67#diff-258035e5968f6bf645400d417f310218d7d9a9a10606a3c34e7f55db193f58f3R3

Instead we're using the rollup replace plugin to read the version at build time and replace `__VERSION__` placeholder with the actual version number.

As a result other packages depending on this one don't need to understand require calls and also we're not exposing the whole package.json file at runtime.

To verify my claim above you can build the package and check the build directory after this change vs before this change.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -3,6 +3,10 @@ import nodeResolve from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
 import terser from '@rollup/plugin-terser';
 import nodePolyfills from 'rollup-plugin-node-polyfills';
+import replace from '@rollup/plugin-replace';
+import fs from 'fs';
+
+const version = JSON.parse(fs.readFileSync('./package.json', 'UTF-8')).version;
 
 export default {
     input: './src/index.ts',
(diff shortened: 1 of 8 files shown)


rollup.config.mjs
```diff
@@ -25,6 +29,10 @@ export default {
         }
     ],
     plugins: [
+        replace({
+            '__VERSION__': version,
+            preventAssignment: true
```
kwasniew, line 34: this is required by the warning from the package. more info here: https://www.npmjs.com/package/@rollup/plugin-replace

src/index.test.ts
```diff
@@ -1383,7 +1381,8 @@ test('Should add `x-unleash` headers', async () => {
         /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
     const expectedHeaders = {
-        'x-unleash-sdk': `unleash-js@${packageJSON.version}`,
+        // will be replaced at build time with the actual version
```
kwasniew, line 1384: tests are working on pre-build version so no replacement

src/version.ts
```diff
@@ -0,0 +1 @@
+export const sdkVersion = `unleash-js@__VERSION__`;
```
kwasniew, line 1: centralizing this to avoid typos in multiple places

thomasheartman (approved):

PR #240, opened 2025-01-16
chore(1-3230): use homebrew version of uuid generation

thomasheartman:
The one from the uuid library relies on an underlying crypto library
which doesn't exist at least in certain GitHub runners and may also
cause issues for react native applications.

This impl sidesteps that issue.

The same uuid generation method that we're cutting out here is also being used in `src/events-handler.ts`. However, because we haven't received any complaints about that, I'll leave it in.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,4 @@
 import { TinyEmitter } from 'tiny-emitter';
-import { v4 as uuidv4 } from 'uuid';
 import Metrics from './metrics';
 import type IStorageProvider from './storage-provider';
 import InMemoryStorageProvider from './storage-provider-inmemory';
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -1,5 +1,4 @@
 import { TinyEmitter } from 'tiny-emitter';
-import { v4 as uuidv4 } from 'uuid';
```
kwasniew, line 2: do we use it for something else? can we remove this from package dependencies?

kwasniew (approved):

src/index.ts
thomasheartman: Good point. I haven't checked, but I'm pretty sure I wasn't the one to add it. Let's see.

thomasheartman: We use it in `src/events-handler.ts` too. We could switch that one for the same impl potentially, but I don't really wanna touch that now if we don't have to?

PR #243, opened 2025-01-30
fix: improve headers parsing

Tymek:
## About the changes
Predictable and testable headers parsing for both getting features and sending metrics.

### DIFF
--- a/src/metrics.test.ts
+++ b/src/metrics.test.ts
@@ -91,7 +91,7 @@ test('should send metrics with custom auth header', async () => {
 
     expect(fetchMock.mock.calls.length).toEqual(1);
     expect(requestBody.headers).toMatchObject({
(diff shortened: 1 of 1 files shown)


src/index.test.ts
```diff
@@ -650,7 +650,7 @@ test('Should abort previous request', async () => {
     client.updateContext({ userId: '456' }); // abort 2
     await client.updateContext({ userId: '789' });
 
-    expect(abortSpy).toBeCalledTimes(2);
```
kwasniew, line 653: how are they different?

kwasniew (approved):

src/index.test.ts
Tymek: one is deprecated. just a rename in a library - no real difference

PR #244, opened 2025-03-26
fix: unstable async bootstrap initialization

Tymek:
## About the changes
Retrieve stored into a temporary variable, instead of overriding it for the duration of a next promise. Avoid race conditions.

Before this change calls to `isEnabled` or `getVariant` could be resolved between stored toggles (empty list) are resolved and bootstrapped toggles are restored. This was causing https://github.com/Unleash/unleash-client-nextjs/issues/106

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -7,6 +7,7 @@ import {
     IConfig,
     IMutableContext,
     IToggle,
+    IVariant,
     InMemoryStorageProvider,
     UnleashClient,
     lastUpdateKey,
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -317,6 +317,51 @@ test('Should bootstrap data when bootstrap is provided', async () => {
...
+        metricsInterval: 0,
+        disableRefresh: true,
+        bootstrapOverride: true,
+        bootstrap,
+        createAbortController: () => null,
+    };
+    const client = new UnleashClient(config);
+
+    for (let i = 0; i < 12; i++) {
+        await true;
```
thomasheartman, line 351: what's this do? awaiting a sync variable seems pointless, but I'm guessing you put it here for some reason?

thomasheartman (approved):
Fix looks sensible to me. I'm not sure I understand the "await true" in the test, though. Is that on purpose? Also: just looping around like this seems like it'd be flaky even if it didn't work. Is there a better way to test this? (might very well not be and not be worth spending time on)

src/index.test.ts
kwasniew: it's repro of this issue: https://github.com/Unleash/unleash-client-nextjs/issues/106

src/index.test.ts
```diff
@@ -317,6 +317,51 @@ test('Should bootstrap data when bootstrap is provided', async () => {
...
+        refreshInterval: 0,
+        metricsInterval: 0,
+        disableRefresh: true,
+        bootstrapOverride: true,
+        bootstrap,
+        createAbortController: () => null,
+    };
+    const client = new UnleashClient(config);
+
+    for (let i = 0; i < 12; i++) {
```
kwasniew, line 350: what's the minumum number of iterations to consistently reproduce it?

kwasniew (approved):

src/index.test.ts
Tymek: ~7

src/index.test.ts
Tymek: Yes. This is important for the bug to occur - it will effectively wrap everything in a promise and add it to a [microtask queue](https://javascript.info/microtask-queue).

PR #246, opened 2025-04-07
fix(metrics): use clearInterval instead of clearTimeout in stop()

### LINKED ISSUE #245: Memory Leak When Using UnleashClient per Request — .stop() Does Not Fully Clean Up
### Describe the bug

When using unleash-proxy-client on the server side (e.g., for SSR or per-request flag fetching), we observed increasing memory usage and event loop handles even though .stop() was called immediately after .start().

After investigation, the issue was resolved by setting disableMetrics: true, suggesting that .stop () does not properly clean up background metrics collection.

Use Case
We are using unleash-proxy-client on the backend to:

- Fetch toggles per request (e.g., server-rendered pages)
- Use short-lived instances of UnleashClient
- Call .start() and .stop() on each request

Observed Behavior

- Memory usage increases with each request
- Open handles (timers) accumulate
- Eventually the process crashes under high load

What Fixed It
Adding this option resolved the issue:
`const unleashClient = new UnleashClient({
  ...
  disableMetrics: true
});
`

After this change:

- Memory usage remained stable
- All client instances cleaned up correctly

- Create a new UnleashClient in a loop or per request
- Call .start(), use the toggles
- Call .stop()
- Watch memory and open handles grow
- Add disableMetrics: true and observe the issue disappear

### Expected behavior

Calling .stop() should:
- Clean up internal timers related to metrics
- Fully dispose of any internal state
- Prevent memory or handle leaks even when disableMetrics is false

### Logs, error output, etc.

```bash

```

### Screenshots

The red line after disabling the metrics

Image

### Additional context

_No response_

### Unleash version

3.6.1

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

unleash-proxy-client 3.6.1

isawalhi:
## About the changes
When using unleash-proxy-client on the server side (e.g., for SSR or per-request flag fetching), we observed increasing memory usage and event loop handles even though .stop() was called immediately after .start().

After investigation, the issue was resolved by setting disableMetrics: true, suggesting that .stop () does not properly clean up background metrics collection.

Closes #245

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -107,7 +107,7 @@ export default class Metrics {
 
     public stop() {
         if (this.timer) {
-            clearTimeout(this.timer);
+            clearInterval(this.timer);
             delete this.timer;
         }
     }


ivarconr (approved):

lo1tuma:
I still have this issue in the latest version 3.7.5

PR #250, opened 2025-04-22
Handle already aborted controllers before re-aborting

ActuallyConnor:
Previously, the code attempted to abort controllers without checking their state, potentially causing issues. This change ensures that controllers are only aborted if they are not already in an aborted state, improving robustness and preventing unnecessary operations.

## About the changes

Closes #

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -545,7 +545,8 @@ export class UnleashClient extends TinyEmitter {
 
     private async fetchToggles() {
         if (this.fetch) {
-            if (this.abortController) {
+            // Check if abortController is already aborted before calling abort
+            if (this.abortController && !this.abortController.signal.aborted) {
                 this.abortController.abort();
             }
             this.abortController = this.createAbortController?.();


Shinypaper:
I've run into a similar issue around this. This fix would be helpful to get in.

Tymek (approved):

kwasniew (approved):

kwasniew:
@ActuallyConnor @Shinypaper what was the impact of calling abort twice? I'd expect the second one to be a no-op.

ActuallyConnor:
> @ActuallyConnor @Shinypaper what was the impact of calling abort twice? I'd expect the second one to be a no-op.

@kwasniew Great question. For our case we noticed it was causing page hang ups occasionally. Our session replay functionality caught the issue and flagged it for me. Just wanted to help out and address our issue simultaneously. 😄

ActuallyConnor:
@kwasniew just curious what the release process would be for this PR once merged? Would it be batched with some other changes or go out as a patch?

kwasniew:
@ActuallyConnor  we can merge it right now and release a patch version

kwasniew:
@ActuallyConnor your change is out as 3.7.6. Thanks for your contribution!

PR #253, opened 2025-06-18
fix: old vs new abort controller

kwasniew:
## About the changes

My hypothesis: in some environments, this.createAbortController might reuse an existing AbortController instance (e.g. a polyfill or a bad custom wrapper).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -553,11 +553,13 @@ export class UnleashClient extends TinyEmitter {
 
     private async fetchToggles() {
         if (this.fetch) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -553,11 +553,13 @@ export class UnleashClient extends TinyEmitter {
 
     private async fetchToggles() {
         if (this.fetch) {
+            const oldController = this.abortController;
+            this.abortController = this.createAbortController?.();
```
kwasniew, line 557: Always create a fresh instance first, then abort the old one, then use the new signal.

PR #254, opened 2025-07-02
fix: make setContextField and removeContextField be async

nitpum:
## About the changes
`setContextField` and `removeContextField` uses `updateToggles` inside like `updateContext` but both `setContextField` and `removeContextField` don't await `updateToggles`. Lead to a case that will get toggle that use old context value.

This fix is add async to both function like `updateContext` does. Now we can await both function

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -373,7 +373,7 @@ export class UnleashClient extends TinyEmitter {
         return { ...this.context };
     }
 
-    public setContextField(field: string, value: string) {
+    public async setContextField(field: string, value: string): Promise<void> {
         if (isDefinedContextField(field)) {
             this.context = { ...this.context, [field]: value };
         } else {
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -373,25 +373,25 @@ export class UnleashClient extends TinyEmitter {
...
+    public async setContextField(field: string, value: string): Promise<void> {
         if (isDefinedContextField(field)) {
             this.context = { ...this.context, [field]: value };
         } else {
             const properties = { ...this.context.properties, [field]: value };
             this.context = { ...this.context, properties };
         }
 
-        this.updateToggles();
+        await this.updateToggles();
```
gastonfournier, line 384: Maybe we should just return `this.updateToggles()` here? 
@kwasniew what do you think about this?

nitpum:
@gastonfournier any update on this?

src/index.ts
kwasniew: In flutter we have Promise<void> equivalent: https://github.com/Unleash/unleash-flutter-sdk/blob/main/lib/unleash_proxy_client_flutter.dart#L350

I think we should incorporate this change as it is now

gastonfournier:
> I think it's a good change and is pretty much aligned with out flutter SDK that I was working on in the past: https://github.com/Unleash/unleash-flutter-sdk/blob/main/lib/unleash_proxy_client_flutter.dart#L350

Thanks! I'm still unsure if this is the way to go... having update toggles as a side effect of setting context fields. Making it event-driven is kind of the same: you can't await updateContext, you'd have to await for an event to arrive. There's an alternative approach in the Java SDK where we allow you to manually call updateToggles which you can await. 

But given @kwasniew is fine with this, I'm also fine with this solution as is.

gastonfournier:
> @gastonfournier any update on this?

Sorry! Summer happens :D Thanks for your patience, I'm merging this now

PR #264, opened 2025-09-04
feat: improve error handling in metrics and fetch toggles

### LINKED ISSUE #263: Expose error type for Unleash SDK error events
### Context

Currently, the Unleash SDK emits error messages without distinguishing their source.
For example:

```
Unleash: unable to send feature metrics TypeError: Failed to fetch
```

or

```
Unleash: unable to fetch feature toggles TypeError: Failed to fetch
```

However, when the SDK propagates the error event, the received object only contains the native exception (TypeError: Failed to fetch), without any indication of whether the failure happened when sending metrics or when fetching feature toggles.

### Problem

In some scenarios, it’s important to distinguish between:

Failure while fetching toggles → critical → requires fallback (e.g., using CDN or local cache) and logging.

Failure while sending metrics → non-critical → can be ignored or logged at a lower priority.

Right now, there’s no reliable way to make this distinction using only the Error object.

### Background

_No response_

### Proposal

Add extra metadata to the emitted error event. For example:

```
emitter.emit('error', {
  type: 'metrics',
  error, // native error object
})

emitter.emit('error', {
  type: 'fetch-toggles',
  error,
})
```

Alternatively, the SDK could attach a property to the error object (error.context = 'metrics').

This would allow consumers to safely filter errors:

```
client.on('error', (err) => {
  if (err.type === 'fetch-toggles') {
    triggerFallback()
    captureMessage('Unleash flags fetch failed', { level: 'warning' })
  }
})
```

gabrielduete:
## About the changes

Closes #263 | https://github.com/Unleash/unleash/issues/10620

### Important files

src/index.ts

## Discussion points

This pull request improves error event handling in the `UnleashClient` class to provide more context about where errors originate. Now, when an error is emitted, it includes a `type` field indicating the source of the error, making it easier to debug and handle errors appropriately.

**Enhanced error event payloads:**

* Updated the `onError` handler in the `Metrics` instantiation to emit errors with a payload containing the error type (`'metrics'`) and the error object, instead of just the error.
* Modified error emission during toggle fetching to include the error type (`'fetch-toggles'`) and the error object, instead of just the error.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -653,15 +653,22 @@ test('Should publish error when initial init fails', (done) => {
         storageProvider,
     };
     const client = new UnleashClient(config);
-    client.start();
+
+    let errorReceived = false;
     client.on(EVENTS.ERROR, (e: any) => {
-        expect(e).toBe(givenError);
-        done();
+        if (!errorReceived) {
+            errorReceived = true;
+            expect(e).toBe(givenError);
+            done();
+        }
     });
+    client.start();
 });
 
 test('Should publish error when fetch fails', (done) => {
-    const givenError = new Error('Error');
+    const givenError = new Error(
+        JSON.stringify({ error: 'Error', type: 'fetch-toggles' })
+    );
 
     jest.spyOn(global.console, 'error').mockImplementation(() => jest.fn());
     fetchMock.mockReject(givenError);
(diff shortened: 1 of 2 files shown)


chriswk (approved):
This makes sense to me. Thank you @gabrielduete . We'll get this merged and published.

PR #280, opened 2026-04-01
fix: prevent double slashes in metrics URL

dharmadeveloper108:
Strips any trailing slash from the base URL before appending `/client/metrics`.

When the configured base URL has a trailing slash (e.g. `http://localhost:3000/api/frontend/`), the SDK constructed URLs like `http://localhost:3000/api/frontend//client/metrics` (spotted while making [this test](https://github.com/Unleash/unleash/blob/main/frontend/src/hooks/useImpactMetrics.test.tsx#L22) in Unleash). 
We never noticed because Unleash runs on Express which normalizes paths by default, but still worth fixing I think.

### DIFF
--- a/src/metrics.test.ts
+++ b/src/metrics.test.ts
@@ -1,6 +1,10 @@
 import { FetchMock } from 'jest-fetch-mock';
 import Metrics from './metrics';
-import { getTypeSafeRequest, parseRequestBodyWithType } from './test';
+import {
+    getTypeSafeRequest,
+    getTypeSafeRequestUrl,
+    parseRequestBodyWithType,
+} from './test';
 import { InMemoryMetricRegistry } from './impact-metrics/metric-types';
 
 jest.useFakeTimers();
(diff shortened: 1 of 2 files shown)


src/metrics.ts
```diff
@@ -149,7 +149,7 @@ export default class Metrics {
     }: { fireAndForget?: boolean } = {}): Promise<void> {
         /* istanbul ignore next if */
 
-        const url = `${this.url}/client/metrics`;
+        const url = `${this.url.toString().replace(/\/$/, '')}/client/metrics`;
```
sighphyre, line 152: I think a regex is both overkill and underkill here 😅. Elsewhere in this SDK we just ingest it into URLs, which validate and clean it up. I think we could do this, which also catches a few cases the regex doesn't catch and matches our existing patterns

```suggestion
        const url = new URL('/client/metrics', this.url).toString();
```

src/metrics.test.ts
```diff
@@ -337,6 +341,49 @@ describe('Custom headers for metrics', () => {
     });
 });
 
+test('should construct correct metrics URL', async () => {
```
sighphyre: Personal taste than anything but I'd throw this test out. Being constructed correctly doesn't tell me what incorrectly actually means and I think the test below kinda expresses your intent perfectly already.

sighphyre (approved):
Nice! Sane and sensible change. I like it

src/metrics.ts
dharmadeveloper108: That's what I went for initially but `new URL('/client/metrics', this.url)` strips the url of the base path 🥲 (so `URL('/client/metrics', 'http://localhost:4242/api/frontend')` -> `http://localhost:4242/client/metrics`, removing `/api/frontend`)

PR #287, opened 2026-06-10
chore: move Node version for tests to more modern 20-26

sighphyre:
Node 16 and 18 are no longer compatible with up to date dev dependencies. Current versions of those dev dependencies have open vulnerabilities, which need to be closed. Node isn't actually a targeted runtime for this SDK, it's just being used to run tests, we can just drop 16+18, it's fine

Skipping the requirements on this - since Node 16 and 18 are required but removed, they won't pass, that's expected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


irstavr (approved):
🆙 🆙  
thanks!

PR #292, opened 2026-07-23
fix: fall back to Math.random() UUID generation when crypto does not exist

koengommers:
## About the changes

After updating to unleash-proxy-client@3.8.0 our React Native app crashed with error `ReferenceError: Property 'crypto' doesn't exist`

In PR https://github.com/Unleash/unleash-js-sdk/pull/284, a change was made that assumes `crypto` exists, which is not the case in a React Native environment.

This fix makes it so that the fallback is used in environments where crypto does not exist.

Closes #


src/uuidv4.ts
```diff
@@ -9,8 +9,8 @@
  * security-sensitive or globally uniqueness-guaranteed tokens.
  */
 export const uuidv4 = (): string => {
-    if (typeof crypto.randomUUID === 'function') {
-        return crypto.randomUUID();
```
gastonfournier, line 13: Thanks, @koengommers! This looks like a good approach.

While reviewing the code, I found an existing pattern in util.ts that I think we should follow here for consistency (it's for sha256 but it also requires crypto library): https://github.com/Unleash/unleash-js-sdk/blob/a805d47b83743805216c44e8d53c55a2aba25cc4/src/util.ts#L42-L62

Could you adapt the PR to use the same pattern? Once that’s done, I’d be happy to accept the contribution. We could make the change ourselves, but it’s always great to keep community contributors involved.

koengommers: Good spot, it should've been `globalThis`, `global` only exists in Node. I've updated it to use globalThis and made the check more similar to the shared snippet.

gastonfournier (approved):
LG! Thanks!

PR #295, opened 2026-08-05
chore: ignore yarn scripts

kwasniew:
## About the changes

  Supply-chain hardening

  Prompted by the recent keyv supply-chain attack (https://www.aikido.dev/blog/keyv-and-friends-compromised-in-npm-supply-chain-attack). We actually have keyv in our dev dependency tree (it comes in via eslint), so this isn't hypothetical for us.

  What's in this PR

  Dependency install scripts are now disabled via ignore-scripts true in .yarnrc. Install scripts are how most of these attacks deliver their payload, and blocking them costs us nothing. I checked, and no package in our tree uses preinstall/install/postinstall today.
  This covers local dev, CI, and the release runner, since they all install from this repo's checkout.

  Publishing still works: npm publish reads .npmrc rather than .yarnrc, so prepack still runs the build. Confirmed with npm pack --dry-run.

  What we already had

  - Frozen lockfile installs in CI and in the release template
  - Trusted-publisher releases since v3.8.2

  What I'm deliberately skipping for now

  Minimum release age protection. Yarn 1 has no support for it, and with the lockfile in place our only exposure is manual dependency bumps anyway. Worth revisiting if we adopt Dependabot/Renovate (both have cooldown settings) or move to pnpm, which has minimumReleaseAge built in.


gastonfournier (approved):
>  I checked, and no package in our tree uses preinstall/install/postinstall today.

I trust you, that's the same thing I was going to check :)

